PR #12, opened 2022-08-02
feat: ID_WEB_Temperatur_TRL_ext support

augi:
This is temperature of the flow-out, measured by an external sensor.

### DIFF
--- a/custom_components/luxtronik/const.py
+++ b/custom_components/luxtronik/const.py
@@ -351,6 +351,12 @@ HEATING_SENSOR_TYPES: tuple[LuxtronikSensorEntityDescription, ...] = (
         sensor_key="calculations.ID_WEB_Temperatur_TRL",
     ),
     LuxtronikSensorEntityDescription(
+        key="flow_out_temperature",
+        # f"{text_flow_out} {text_temp} ({text_external})",
+        icon="mdi:waves-arrow-right",
+        sensor_key="calculations.ID_WEB_Temperatur_TRL_ext",
+    ),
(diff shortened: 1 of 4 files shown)


custom_components/luxtronik/const.py
```diff
@@ -350,6 +350,12 @@ class LuxMode(Enum):
         icon="mdi:waves-arrow-right",
         sensor_key="calculations.ID_WEB_Temperatur_TRL",
     ),
+    LuxtronikSensorEntityDescription(
+        key="flow_out_temperature",
```
BenPru, line 354: This key is already used for "calculations.ID_WEB_Temperatur_TRL".
Please change it.

custom_components/luxtronik/sensor.py
```diff
@@ -351,6 +352,16 @@ async def async_setup_entry(
                 "mdi:waves-arrow-right",
                 entity_category=None,
             ),
+            LuxtronikSensor(
+                hass,
+                luxtronik,
+                device_info_heating,
+                "calculations.ID_WEB_Temperatur_TRL_ext",
+                "flow_out_temperature",
```
BenPru, line 360: This key is already used for "calculations.ID_WEB_Temperatur_TRL".
Please change it.

BenPru (changes requested):
The key "flow_out_temperature" is already used for "calculations.ID_WEB_Temperatur_TRL".
Please change it.

BenPru (approved):

Issue #17, opened 2022-09-27
Heating climate without data


augi:
The `climate.luxtronik2_domestic_water` entity shows this nice graph ❤️ 
![image](https://user-images.githubusercontent.com/273052/192508339-dd468f05-8bc2-4958-8dbe-f5c11d2f0ed8.png)

Unfortunately, the `climate.luxtronik2_heating` entity looks like being without data 😢 

![image](https://user-images.githubusercontent.com/273052/192508689-a98f7f36-4bf1-4e5d-a148-5c00472a315b.png)

Don't you please know what could be wrong? Also, please note that all the statuses (e.g. `sensor.luxtronik2_status`) and temperatures get the right data.

scaronni:
Just a note, the above merge request does not fix it for me, I have roughly the same issue with something weird on the heating target temperature after updating the integration:

![image](https://user-images.githubusercontent.com/1983966/201695975-92d47025-b9a6-4ee0-b7b8-1594235725fe.png)

scaronni:
I've just noticed that now instead of picking up the set temperature (in my case 21 degrees) it now picks the temperature correction. This is visibile also directly from the entity `climate.luxtronik_heating` which no longer displays the wanted temperature:

![image](https://user-images.githubusercontent.com/1983966/201715104-3b1aeac4-074c-47be-bd43-40845695e474.png)

rhammen:
> I've just noticed that now instead of picking up the set temperature (in my case 21 degrees) it now picks the temperature correction. This is visibile also directly from the entity `climate.luxtronik_heating` which no longer displays the wanted temperature:
> 
> ![image](https://user-images.githubusercontent.com/1983966/201715104-3b1aeac4-074c-47be-bd43-40845695e474.png)

Previously, you could change the heating target temperature with the climate control, but it didn't actually change anything on the Luxtronik. In order to get to a real climate-control, I needed to pick a sensor that controls the temperature of the heatpump. For me, the main sensor to choose was therefore the temperature correction.
Do you know another sensor that can be used to control the heating behavior of the heatpump, but better 'reflects' the heating target temperature?

rhammen:
> Just a note, the above merge request does not fix it for me
> 
> ![image](https://user-images.githubusercontent.com/1983966/201695975-92d47025-b9a6-4ee0-b7b8-1594235725fe.png)

The "heating current temperature" will show the value of the sensor that you configured. 
[...]

Issue #15, opened 2022-09-19
Room temperature sensor not available


Laars123:
My luxtronik/alpha innotec installation includes a room temperature sensor.
However, it doesn't show up in the integration. Can this be added?
It does show up in de webserver heatpump controller though.
Thanks!

Note: Auto discovery didn't work in my case.

rhammen:
I'd also like to have this :-)
parameter: ID_WEB_RBE_RT_Ist

rhammen:
I was able to add the Room Temperature by adding the following around line 400 of sensor.py:
            LuxtronikSensor(
                hass,
                luxtronik,
                device_info_heating,
                "calculations.ID_WEB_RBE_RT_Ist",
                "room_temperature",
                f"Room {text_temp}",
                entity_category=None,
            ),

Additionally, I added a sensor for the Pumpfrequency at around line 325 in sensor.py:
        LuxtronikSensor(
            hass,
            luxtronik,
            device_info,
            "calculations.ID_WEB_Freq_VD",
            "pump frequency",
            f"Pump Frequency",
            entity_category=None,
            icon="mdi:sine-wave",
            unit_of_measurement='Hz'
        ),

Kars-de-Jong:
For the climate entity, you can add this yourself using the configuration dialog. Just enter “calculations.ID_WEB_RBE_RT_Ist” in the “Home Assistant sensor id for the current indoor temperature” text field.

@rhammen that is not the pump frequency, it’s the compressor frequency.

You can add extra sensors yourself using the “luxtronic2” sensor platform:

```yaml
sensors:
  - platform: luxtronik2
    sensors:
      - group: calculations
        id: ID_WEB_Temperatur_THG
        friendly_name: Temperatuur heetgas
        icon: mdi:thermometer
        state_class: measurement
      - group: calculations
        id: Heat_Output
        friendly_name: Vermogen
        icon: mdi:flash
        state_class: measurement
        unit_of_measurement: W
```
etc.

BenPru:
I have no luxtronik hardware thermostat.
Is it possible to write the ID_WEB_RBE_RT_Ist from ha to luxtronik?
ID_WEB_RBE_RT_Ist is a calculation and readonly, but perhaps we can find a parameter for this?

rhammen:
[...]

Issue #6, opened 2021-11-10
Climate thermostat cooling


BenPru:
Creating a home assistant climate thermostat for the cooling

Current state: open

Todos:
- [ ] Sensors:
  - status_sensor='calculations.ID_WEB_WP_BZ_akt',
  - current_temperature_sensor=Select a sensor from ha???,
  - target_temperature_sensor='parameters.ID_Sollwert_KuCft2_akt'?,
  - target_temperature_sensor_write='parameters.ID_Sollwert_KuCft2_akt'?,
  - heater_sensor='parameters.ID_Einst_BA_Kuehl_akt'?

rhammen:
In the proposed changes I added sensors for both:
- "parameters.ID_Sollwert_KuCft2_akt" --> = target water temperature used during cooling
- "parameters.ID_Einst_KuehlFreig_akt" --> = outdoor temperature threshold. Cooling starts only if T_outside > (this parameter).
In my view, you cannot really use ID_Sollwert_KuCft2_akt to steer on, because of the risk of condensation. So I have it fixed at 18 C.
Therefore I choose to use ID_Einst_KuehlFreig_akt as target_temperature_sensor: it defines the minimal outdoor temperature before cooling starts.
I can imagine that opinions on this subject may differ, so maybe we should add the option to choose which sensor to control.

Kars-de-Jong:
I commented on your latest PR, this implementation is too simple (it's not always mixer circuit 2 that is used for cooling).

rhammen:
I'd say we now have a working thermostat. The thermostat allows to set the outdoor temperature above which cooling will start.
I think some documentation - explaining the usage - may still be helpful/needed.
What else do we want/need as part of this issue?

AJediIAm:
Closing the issue since it was implemented some time ago. If more functionality is needed, a new (up to date) feature can be created.

PR #19, opened 2022-11-13
Several enhancements

rhammen:
- Added sensors for Compressor Frequency and Room temperature sensor (issue #15  )
- Enhanced the cooling climate control (related to #6 )
- Heating climate control now controls the temperature correction like in the Alpha Control App
- Looks like it fixes issue #17


BenPru (approved):
Nice work, thanks

PR #26, opened 2022-11-17
Enhanced manufacturer detection and changed cooling present detection

rhammen:


custom_components/luxtronik/const.py
```diff
@@ -190,7 +190,7 @@ class LuxMode(Enum):
 
 
 # region Luxtronik Sensor ids
-LUX_SENSOR_DETECT_COOLING: Final = "calculations.ID_WEB_FreigabKuehl"
+LUX_SENSOR_DETECT_COOLING: Final = "parameters.ID_Sollwert_KuCft2_akt"
```
Kars-de-Jong, line 193: This is the cooling setpoint for mixer circuit 2. That is only valid if the function of mixer circuit 2 is set to "Cool".
Are you sure this value is 0 when no cooling option is installed? Because for instance on my system, the setting ID_Sollwert_KuCft1_akt (cooling setpoint for mixer circuit 1) is 19.0, and I don't have mixer 1 in my setup.

rhammen: Thanks for the proposed detection method, I agree this should work more reliable. 
I updated the code to implement the proposed detection method
In a similar way I added detecting of the Cooling Target temperature sensor

rhammen:
Updated the detection method based on the suggestion of @Kars-de-Jong

BenPru:
@Kars-de-Jong Is the newer version correct?
I have no cooling and I can't check it.

custom_components/luxtronik/__init__.py
```diff
@@ -145,6 +147,14 @@ def setup_internal(hass, data, conf):
         manufacturer=get_manufacturer_by_model(model),
         model=model,
     )
+    
+    MK1 = luxtronik.get_value('parameters.ID_Einst_MK1Typ_akt')
+    MK2 = luxtronik.get_value('parameters.ID_Einst_MK2Typ_akt')
+    MK3 = luxtronik.get_value('parameters.ID_Einst_MK3Typ_akt')
+    CoolingPresent = ((MK1 in [3,4]) | (MK2 in [3,4]) | (MK3 in [4]))
```
Kars-de-Jong, line 154: Magic numbers, I think you should define constants for these, or better, add support for these to the python-luxtronik library.

Kars-de-Jong: Also, create helper methods for this because this pattern is repeated for the detection of the setpoint parameter.

rhammen: Detection functions created in luxtronik_device.py

BenPru (approved):
Great work, thank you very much.

PR #33, opened 2022-11-21
Added Heat_output (Leistung Ist) to Sensors

Crashman1983:


custom_components/luxtronik/const.py
```diff
@@ -371,6 +372,17 @@ class LuxMkTypes(Enum):
...
         sensor_key="calculations.ID_WEB_WMZ_Seit",
     ),
+        LuxtronikSensorEntityDescription(
+        key="current_heat_output",
+        # name=f"{text_heat_amount_counter}",
+        icon="mdi:lightning-bolt-circle",
+        device_class=DEVICE_CLASS_ENERGY,
+        state_class=STATE_CLASS_MEASUREMENT,
+        unit_of_measurement=POWER_WATT,
+        #native_unit_of_measurement="W",
```
Kars-de-Jong, line 382: I believe Home Assistant sensors are supposed to only use `native_unit_of_measurement` these days.

custom_components/luxtronik/const.py
```diff
@@ -371,6 +372,17 @@ class LuxMkTypes(Enum):
         entity_category=EntityCategory.DIAGNOSTIC,
         sensor_key="calculations.ID_WEB_WMZ_Seit",
     ),
+        LuxtronikSensorEntityDescription(
+        key="current_heat_output",
+        # name=f"{text_heat_amount_counter}",
+        icon="mdi:lightning-bolt-circle",
+        device_class=DEVICE_CLASS_ENERGY,
```
Kars-de-Jong, line 379: This should be `DEVICE_CLASS_POWER`.

custom_components/luxtronik/sensor.py
```diff
@@ -300,6 +301,19 @@ async def async_setup_entry(
...
         ),
+        LuxtronikSensor(
+            hass,
+            luxtronik,
+            device_info,
+            sensor_key="calculations.Heat_Output",
+            unique_id="current_heat_output",
+            name=f"{text_current_heat_output}",
+            icon="mdi:lightning-bolt-circle",
+            device_class=DEVICE_CLASS_ENERGY,
```
Kars-de-Jong, line 312: This should be `DEVICE_CLASS_POWER`.

custom_components/luxtronik/sensor.py
```diff
@@ -300,6 +301,19 @@ async def async_setup_entry(
...
+            state_class=STATE_CLASS_MEASUREMENT,
+            unit_of_measurement=POWER_WATT,
```
Kars-de-Jong, line 314: Not sure whether this is needed, I think it is the default for `DEVICE_CLASS_POWER`.

custom_components/luxtronik/const.py
```diff
@@ -371,6 +372,17 @@ class LuxMkTypes(Enum):
         entity_category=EntityCategory.DIAGNOSTIC,
         sensor_key="calculations.ID_WEB_WMZ_Seit",
     ),
+        LuxtronikSensorEntityDescription(
+        key="current_heat_output",
+        # name=f"{text_heat_amount_counter}",
```
Kars-de-Jong, line 377: `name=f"{text_current_heat_output}"`, but it's commented out, doe we still have to add this for new sensors?

Kars-de-Jong (changes requested):
These files should not be committed, `.DS_Store` should be added to the `.gitignore` file:

`.DS_Store`

This will ignore it for all directories.

custom_components/luxtronik/const.py
Crashman1983: doesn't work and is commented out for other sensors as well.

Crashman1983:
All comments fixed, besides naming

custom_components/luxtronik/sensor.py
```diff
@@ -300,6 +301,19 @@ async def async_setup_entry(
             unit_of_measurement=ENERGY_KILO_WATT_HOUR,
             entity_category=EntityCategory.DIAGNOSTIC,
         ),
+        LuxtronikSensor(
```
BenPru, line 304: I prefer to check if the sensor is present. My value is always "None".
e.g.:
has_heat_output = luxtronik.get_value("calculations.Heat_Output") is not None
if has_heat_output:

Crashman1983: sorry, just copied the pattern from other sensors. Where should I place the "if"?

Crashman1983: Think I did it :-)

custom_components/luxtronik/const.py
Kars-de-Jong: OK, but the commented out text is wrong.

Kars-de-Jong (approved):

PR #34, opened 2022-11-21
Correct pump names

BenPru:


custom_components/luxtronik/binary_sensor.py
```diff
@@ -210,12 +211,21 @@ async def async_setup_entry(
...
+            name=text_heating_main_circulation_pump,
+            icon="mdi:pump",
+            device_class=DEVICE_CLASS_RUNNING,
+        ),        
+        LuxtronikBinarySensor(
+            hass=hass,
+            luxtronik=luxtronik,
+            deviceInfo=deviceInfo,
+            sensor_key='calculations.ID_WEB_VBOout',
+            unique_id="fan_brine_or_solar_heat_source_pump",
```
Kars-de-Jong, line 224: It should not be "solar". The translation from German may be a bit problematic here.
"Ventilator, Brunnen- oder Soleumwälzpumpe"
Ventilator -> Fan
Brunnen -> Well
Sole -> Brine

PR #57, opened 2022-12-20
Change minimum value of heating_room_temperature_impact_factor to 0

Kars-de-Jong:
You should be able to set it to 0 to completely disable the RBE. This can be useful in an automation which detects that the window is open in the room with the RBE.

The heatpump menu also allows this.


BenPru (approved):
Thanks

PR #73, opened 2023-01-22
Update icons for low and high pressure sensors

Kars-de-Jong:
Without this change, the pressure sensors are shown using a thermometer icon.


BenPru:
@Kars-de-Jong 
Thanks for your pr.
I have changed it to strings and changed the high icon to gauge-full.
Are you okay with this?

Kars-de-Jong:
@BenPru Ah, sorry about the strings. That’s what you get for editing online outside an IDE.

I used gauge because it is the exact opposite of gauge-low. The scale shows 4 dots, gauge-low is at the second, gauge is at the third.

The opposite of gauge-full (which is full scale) is gauge-empty, but I found those too extreme.

BenPru (approved):

PR #74, opened 2023-01-22
Heating Thermostat v0.5

rhammen:
[BETA]
First version of the Heating Thermostat (#3) .
The Heathing Thermostat must be enabled by using  "Configure" on the integration. (And can also be disabled again in the same way).
The Heating Thermostat controls the Luxtronik heating target_correction, as follows:
target_correction = 100% * (T_Thermostat - T_Room).
T_Thermostat is the temperature you set on the Thermostat.
T_Room is the room temperature, based on the sensor you defined in "Configure".

Example:
- T_Thermostat = 21.0 C
- T_Room = 20.0 C
--> target_correction = +1.0 C
This means that the heatpump will set the Return Flow target to a +1.0 C (in an attempt to decrease the temperature difference quicker).

Some additional notes:
- The target_correction will be set in multiples of 0.5 C
- During HA start, the T_Thermostat will be set based on the current T_room and current target_correction
- It may take 60 seconds after start for the Heating Thermostat to show the current values for the first time.
- In case the Heating Thermostat is disabled, the Thermostat will show the current room temperature, but it will not now the target temperature. Changing the thermostat value is possible will not result in anything.
- In case an RBE (room wall thermostat) is configured, the RBE will already act as a thermostat. The HA thermostat will now act on T_Thermostat(HA) - T_Thermostat(RBE).

This feature BETA, please report any issues you may observe or questions you may have.
The Logs currently contain quite some logging on the thermostat.
In case of issues, it helps to provide the relevant part of the logs.


BenPru:
@rhammen Thanks for your pr. I try to test it locally with my heatpump.
I'm currently recreating this project. The current code is available in the branch [core-preview](https://github.com/BenPru/luxtronik/tree/core-preview).
At the moment I have a rudimentary climate thermostat only for heating there. The domestic water is moved to the platform water_heater. And cooling is currently not implemented.
Perhaps we should merge your changes in this branch.
And another idea: Do we need an extra theromstat for cooling? Or should it be handled with the same heating thermostat?

rhammen:
> And another idea: Do we need an extra thermostat for cooling? Or should it be handled with the same heating thermostat?

I can imagine one combined heating/cooling thermostat that works like this:
```
 IF sensor.luxtronik2_outdoor_temperature_average < number.luxtronik2_heating_threshold_temperature: 
     <Thermostat in Heating mode>: 
    Thermostat controls number.luxtronik2_heating_target_correction based on (<Thermostat target> - room temperature )
ELSE
    <Thermostat in Cooling mode>:
    Thermostat controls directly number.luxtronik2_cooling_threshold_temperature . 
   (This parameter is the outdoor temperature above which the heat-pump will start cooling) 
```

BenPru:
> controls directly number.luxtronik2_cooling_threshold_temperature

I think `cooling_threshold_temperature` is a config setting value and should not set via thermostat. I think the temperature param to set is `ID_Sollwert_Kuehl1_akt`.
Perhaps we can use the middle of `heating_threshold_temperature` + `cooling_threshold_temperature` to detect heating or cooling "season". If we use only `heating_threshold_temperature` the climate thermostat perhaps changes very often in the heating season to cooling.
We can check at midnight which mode should be active and change the displayed name and available modes.
Heating:
* hvac_modes=`[HVACMode.HEAT, HVACMode.OFF]`,
* preset_modes=`[PRESET_NONE, PRESET_AWAY, PRESET_BOOST]`,
* supported_features=`ClimateEntityFeature.AUX_HEAT | ClimateEntityFeature.PRESET_MODE | ClimateEntityFeature.TARGET_TEMPERATURE`

Cooling:
* hvac_modes=`[HVACMode.COOLING, HVACMode.OFF]`,
* preset_modes=`[]`,
* supported_features=`ClimateEntityFeature.TARGET_TEMPERATURE`


```
IF `has_cooling` AND `outdoor_temperature_average` > (`heating_threshold_temperature` + `cooling_threshold_temperature`) / 2: 
    <Thermostat in Cooling mode>:
    Thermostat controls directly parameters.ID_Sollwert_Kuehl1_akt. 
ELSE
     <Thermostat in Heating mode>: 
    Thermostat controls number.luxtronik2_heating_target_correction based on (<Thermostat target> - room temperature )
```

rhammen:
Sorry, but I really think the cooling thermostat should set the cooling_threshold_temperature: "parameters.ID_Einst_KuehlFreig_akt". This is the main parameter that determines if the heatpump should start cooling or not.
At my heatpump ID_Sollwert_Kuehl1_akt = 0; it could be that this is a binary parameter that defines if cooling may start or not, but I think this is not a temperature setting as you suggest.
I don't think there is a need to make it more complex than it needs to be. The heatpump will start heating when the outdoor (average?) temperature is below heating_threshold_temperature . Similarly, the heatpump will start cooling when the outdoor (average?) temperature is above cooling_threshold_temperature . 
Between cooling_threshold_temperature  and heating_threshold_temperature  will neither cool not heat.
So I'd say just checking that cooling_threshold_temperature > heating_threshold_temperature should be enough?

The main problem/challenge with cobining the heating and cooling thermostat is that the PRESET_modes and their target (cooling or heating) change.
If you combine them in 1 thermostat, you basically can't set/access the heating modes in summer, and the cooling modes in winter. This may confuse users, especially in spring and fall...

In the Alpha Control App their are 3 entries year-round: heating, domestic water, and cooling.
Why do we want to do it differently?
So this brings us back to the question: should we really combine them into 1 thermostat?
Is the advantage (1 thermostat) worth the drawbacks (less logical behaviour) ?

BenPru:
> So I'd say just checking that cooling_threshold_temperature > heating_threshold_temperature should be enough?

Its okay for me. Unfortunately, my heat pump has no cooling and so I'm not that deep into the cooling matter.

> In the Alpha Control App there are 3 entries year-round: heating, domestic water, and cooling.
> Why do we want to do it differently?

Who uses this app? It's not the best.
When do you need heating and cooling at the same time?
Have anyone two RBEs - one for heating and one for cooling?
The HA climate thermostat can heat and cool. So why sould we not use it?

AJediIAm:
I looked at this feature and I think it does not work well with my heatpump. 

When I change the temperature from 20C to 21C using the room thermostat, this feature will increase the correction temperature to +1C, causing the heatpump to overshoot the 21C. 

These kind of corrections should be made through an automation template supplied seperatly. Having this part of an integration is against the design filosofie of home assistant.

AJediIAm:
This integration should be used instead: https://github.com/custom-components/climate.programmable_thermostat

rhammen:
> I looked at this feature and I think it does not work well with my heatpump.
> 
> When I change the temperature from 20C to 21C using the room thermostat, this feature will increase the correction temperature to +1C, causing the heatpump to overshoot the 21C.

Do I understand it correctly that you installed the changed files in this PR, and you observed the room temperature overshooting the temperature you set with the thermostat? This is not intended: when the room temperature gets closer to the target temperature (21C), the correction should have been reduced. E.g. at room = 20.5 C, the correction should have been changed to 0.5 C. And at Room = 21C, the correction should have been set to 0.
Can you share the full HA logs when you experienced this, so I can have a look to see if I can understand what happened?

BenPru:
I tested this pull request. And it didn't work for me as expected. But I'm afraid my automation interfered and changed the target correction.
Unfortunately, my test is not conclusive.
However, I noticed that the thermostat (heating and hot water) shows hvac_action heating although there is no heating.

AJediIAm:
> > I looked at this feature and I think it does not work well with my heatpump.
> > When I change the temperature from 20C to 21C using the room thermostat, this feature will increase the correction temperature to +1C, causing the heatpump to overshoot the 21C.
> 
> Do I understand it correctly that you installed the changed files in this PR, and you observed the room temperature overshooting the temperature you set with the thermostat? This is not intended: when the room temperature gets closer to the target temperature (21C), the correction should have been reduced. E.g. at room = 20.5 C, the correction should have been changed to 0.5 C. And at Room = 21C, the correction should have been set to 0. Can you share the full HA logs when you experienced this, so I can have a look to see if I can understand what happened?

I had something similar implemented through an automation to adjust the room temperature, but the temperature became very unstable. During cooking, more cold air would come in die to the extra ventilation and the room temperature would drop half a degree. As a result, the floor heating would get a bit hotter. This heat would stay in the house for the rest of the evening.
I have a well insulated house with a relatively large heating surface, so it takes a long time to cool down.

There is a reason the heatpump does not take the room temperature into account to adjust the heating. Once the heating pattern is defined correctly, no adjustments are needed.

BenPru:
> There is a reason the heatpump does not take the room temperature into account to adjust the heating. Once the heating pattern is defined correctly, no adjustments are needed.

I use two indoor temperatures. The first is the average temp from all room thermostats. And yes if the sun shines in the living room this temperature rises fast and is not good to use for heating.
The second and for me the primary heating temperature is the air outgoing temperature from my ventilation system. It is slow and has 0.5 C° steps. But it is stable and a very good indicator.

PR #78, opened 2023-01-25
Optimize detect_solar_present

BenPru:


rhammen:
For me, this change did not fix the problem. Then new code in lines 159 and 162 still checks with an OR for these visibilities.

PR #81, opened 2023-01-29
Another attempt to fix the solar detection

rhammen:
2023.01.29 still added solar sensors to my setup, where I do not have a solar collector :-(

The problem in my view is that 
1) the solar detection from luxtronik_device.py was not used (disabled) in sensor.py
2) in my heatpump:
LUX_DETECT_SOLAR_SENSOR = 0
ID_Visi_Temp_Solarkoll =1
ID_Visi_Temp_Solarsp = 1
ID_WEB_Temperatur_TSK = 5 
ID_WEB_Temperatur_TSS = 150

I tried fixing the problem for my setup without breaking it for others...


BenPru:
I think this is good.
Please note that I will always generate some of the sensors and in the future all of them. I just set the default enabled status to true or false. However, if this status was (wrongly) set by a previous version, the original status may be retained.

PR #106, opened 2023-06-30
Add texts for switchoff_reason Schaltspielsperre/Switching cycle lock

macrojames:
Just encountered Switchoff: 18 and looked at the heatpump.


BenPru (approved):
Thanks

PR #109, opened 2023-07-29
Create repository.yaml

dinkelnaar:
Tryto get integrated in homeassist. Manual says this file needs to be there.


AJediIAm:
Wouldn't it be better to list BenPru as the maintainer since he's the one who made all of this?

BenPru (approved):

PR #118, opened 2023-09-08
Reflect "passive cooling" in sensor.luxtronik2_status

rhammen:
Update sensor.luxtronik2_status , so that it shows 'Cooling' when the system is in passive cooling mode.

Since (as far as I know) there is no explicit reporting of the passive cooling status by the luxtronik device, I had to do a bit of a hack to detect the passive cooling status. It is not 100% accurate, but it works pretty well for me. Note: the Cooling Device does not yet report it is cooling!

Please report if it works for you if you have a passive cooling system. 
Also in case you encounter problems, report them  here

### DIFF
--- a/custom_components/luxtronik/sensor.py
+++ b/custom_components/luxtronik/sensor.py
@@ -34,6 +34,7 @@ from .const import (ATTR_EXTRA_STATE_ATTRIBUTE_LUXTRONIK_KEY, ATTR_STATUS_TEXT,
                     LUX_STATUS1_HEATPUMP_SHUTDOWN, LUX_STATUS1_WORKAROUND,
                     LUX_STATUS3_WORKAROUND, LUX_STATUS_DOMESTIC_WATER,
                     LUX_STATUS_EVU, LUX_STATUS_HEATING, LUX_STATUS_NO_REQUEST,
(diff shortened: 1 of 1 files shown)


AJediIAm:
Tested on a Alpha Innotec WZSV 62 K3M
It seems to work as described with no errors.

custom_components/luxtronik/sensor.py
```diff
@@ -915,6 +916,22 @@ def native_value(self):  # -> float | int | None:
                 return LUX_STATUS1_HEATPUMP_SHUTDOWN
             # endregion Workaround Luxtronik Bug: Line 1 shows 'heatpump coming' on shutdown!
 
+        
+        # workaround to detect (passive) cooling active
+        if (self._sensor_key == LUX_SENSOR_STATUS) and (value == LUX_STATUS_NO_REQUEST):     
+            if self._luxtronik.detect_cooling_present():
+                T_in       = self._luxtronik.get_value("calculations.ID_WEB_Temperatur_TVL")
```
BenPru, line 923: The var names break the HA Style guidelines.
https://developers.home-assistant.io/docs/development_guidelines/
Pleas use
temp_out
and so on.

BenPru (approved):

PR #119, opened 2023-09-09
Fix Capitalization and update icon

AJediIAm:
Fix Capitalization Of Status Names
Change cooling icon for better fit with water-water heat pumps.


BenPru (changes requested):
This doesn't work with the translation files.
Please see
translations/sensor.*.json

Issue #130, opened 2023-09-17
2023.9.12-Alpha-Core-Preview: wrong state-class for "energy" device-class


AJediIAm:
I got the following warning:
![image](https://github.com/BenPru/luxtronik/assets/5879533/5808c2c4-3c4e-4d9e-9df0-3ec1151b1dc9)

Looking into https://developers.home-assistant.io/docs/core/entity/sensor/, I have some doubts about the energy device class. 
Although it is energy output, it is heat energy and not electrical energy which often confuses people. 
Changing the state class from "measurement" to "total increasing" is the easiest fix. 

![image](https://github.com/BenPru/luxtronik/assets/5879533/adb36158-5691-42e4-a21c-f0582bfccaf9)

AJediIAm:
I think you are correct. I found a reference to parameter 257 here which agrees with power as the class: https://community.openhab.org/t/luxtronikheatpump-finding-the-currently-drawn-power/121605/10

BenPru:
Is this fixed?

rhammen:
> Is this fixed?

I did not see these errors anymore, so looks fixed to me.

PR #131, opened 2023-09-17
Update sensor state class in sensor_entities_predefined.py

AJediIAm:
Fix for issue #130


custom_components/luxtronik/sensor_entities_predefined.py
```diff
@@ -330,10 +330,10 @@
         key=SensorKey.CURRENT_HEAT_OUTPUT,
         luxtronik_key=LC.C0257_CURRENT_HEAT_OUTPUT,
         icon="mdi:lightning-bolt-circle",
-        state_class=SensorStateClass.MEASUREMENT,
+        state_class=SensorStateClass.TOTAL_INCREASING,
         device_class=SensorDeviceClass.ENERGY,
         entity_category=EntityCategory.DIAGNOSTIC,
-        native_unit_of_measurement=UnitOfPower.WATT,
+        native_unit_of_measurement=UnitOfEnergy.KILO_WATT_HOUR,
```
BenPru, line 336: Can we simple change from Watt to KiloWatt?
My heatpump doesn't have this sensor. But I think if luxtronik sends the value in Watt we have to set a factor: 0.001

AJediIAm: Yes

Watt is not a valid unit of measurement for Energy. It should be Wh or kWh.
image
In this case, it is KwH, just like the other sensors with SensorDeviceClass.ENERGY

The sensors are currently 0 due to another bug (probably related to the external temperature).

Issue #137, opened 2023-09-19
Lot of duplicate (one working, the other not working) sensors after update to AlphaPreview 2023.09.19


rhammen:
I do see a lot of duplicate sensors in the integration after updating to AlphaPreview 2023.09.19

One sensor is greyed-out, and "This entity is no longer being provided by the luxtronik2 integration."
The other sensor (with the same name) is then  the new one.

Below a list of all Registered sensors, from the logs
It looks like the problem occurs for all (new) sensors that have a _2 at the end of their entity.
So instead of updating the existing sensors the first time the AlphaPreview is run, a new sensor with _2 at the end is added, and the old sensor is not registered anymore.

2023-09-20 00:10:57.810 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new sensor.luxtronik2 entity: sensor.luxtronik2_compressor1_impulses
2023-09-20 00:10:57.811 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new sensor.luxtronik2 entity: sensor.luxtronik2_compressor1_operation_hours
2023-09-20 00:10:57.813 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new sensor.luxtronik2 entity: sensor.luxtronik2_compressor2_impulses
2023-09-20 00:10:57.814 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new sensor.luxtronik2 entity: sensor.luxtronik2_compressor2_operation_hours
2023-09-20 00:10:57.817 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new sensor.luxtronik2 entity: sensor.luxtronik2_heat_amount_flow_rate
2023-09-20 00:10:57.825 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new sensor.luxtronik2 entity: sensor.luxtronik2_heat_source_flow_rate
2023-09-20 00:10:57.834 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new sensor.luxtronik2 entity: sensor.luxtronik2_additional_heat_generator_operation_hours
2023-09-20 00:10:57.836 INFO (MainThread) [homeassistant.helpers.entity_registry] Registered new sensor.luxtronik2 entity: sensor.luxtronik2_current_heat_output
[...]

PR #145, opened 2023-10-10
Fix for duplicate _2 sensors

rhammen:
Fix for duplicate _2 sensors #137


BenPru (approved):
LGT, but I think we should add a new upgrade step and data version 6.

Issue #149, opened 2023-10-12
[2023.10.12-Beta] climate and water_heater could not be set up


ChrisMisker:
Hi,
I get this error in the notification:
<img width="477" alt="Scherm­afbeelding 2023-10-12 om 07 02 09" src="https://github.com/BenPru/luxtronik/assets/37826263/d9e04bee-e040-4c8a-9f4c-e48859481781">

When inspecting the logs, I get these errors:
```
Logger: homeassistant.setup
Source: setup.py:359
First occurred: 06:58:16 (2 occurrences)
Last logged: 06:58:16

Unable to prepare setup for platform luxtronik2.climate: Platform not found (Exception importing custom_components.luxtronik2.climate).
Unable to prepare setup for platform luxtronik2.water_heater: Platform not found (Exception importing custom_components.luxtronik2.water_heater).
```

```
Logger: homeassistant.loader
Source: loader.py:838
First occurred: 06:58:16 (1 occurrences)
Last logged: 06:58:16

Unexpected exception importing platform custom_components.luxtronik2.water_heater
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/loader.py", line 838, in get_platform
    cache[full_name] = self._import_platform(platform_name)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/loader.py", line 855, in _import_platform
    return importlib.import_module(f"{self.pkg_path}.{platform_name}")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
[...]

PR #150, opened 2023-10-12
Fix for #149

rhammen:
Fix for #149


BenPru (approved):
unit_of_measurement is the default entity description property.
Approve without depper check.

rhammen:
FYI: The [ClimateEntity](https://developers.home-assistant.io/docs/core/entity/climate/) and the Water Heater Entity do not specify (do not allow) a "unit_of_measurement " property. Both do however allow the "temperature_unit" property, the Climate Entity even requires it. That is why I changed "unit_of_measurement" --> "temperature_unit"
This fixed a problem visible in the UI. For example look at https://user-images.githubusercontent.com/32298537/268449679-0509be37-c97e-4ccc-9015-b077e4ed2990.png
It displayed "Off oC". 
This change in the specified property removes the oC from the UI, so the mode correctly shows "Off" without any unit.

PR #154, opened 2023-10-15
Several fixes/improvements

rhammen:
- add missing cooling sensors, incl. translations
- fix correction calculation in heating thermostat
- fix solar detection
- fix cooling thermostat showing heating
-


custom_components/luxtronik/climate.py
```diff
@@ -53,14 +53,25 @@
 MIN_TEMPERATURE = 8
 MAX_TEMPERATURE = 28
 
-HVAC_ACTION_MAPPING: dict[str, str] = {
+HVAC_ACTION_MAPPING_HEAT: dict[str, str] = {
     LuxOperationMode.heating.value: HVACAction.HEATING.value,
-    LuxOperationMode.domestic_water.value: HVACAction.HEATING.value,
+    LuxOperationMode.domestic_water.value: HVACAction.IDLE.value,
```
BenPru, line 58: When hot water is produced, the underfloor heating also receives heat.
So the heating thermostat is active if domestic_water is active.

custom_components/luxtronik/number_entities_predefined.py
```diff
@@ -463,15 +463,88 @@
...
+        key=SensorKey.COOLING_OUTDOOR_TEMP_THRESHOLD,
+        luxtronik_key=LP.P0110_COOLING_OUTDOOR_TEMP_THRESHOLD,
+        device_key=DeviceKey.cooling,
+        icon='mdi:sun-thermometer',
+        device_class=SensorDeviceClass.TEMPERATURE,
+        native_unit_of_measurement=UnitOfTemperature.CELSIUS,
+        native_min_value=18.0,
+        native_max_value=30.0,
+        native_step=0.5,
+        #entity_category=EntityCategory.CONFIG,
```
BenPru, line 476: Why not in config?

custom_components/luxtronik/translations/de.json
```diff
@@ -422,6 +440,9 @@
             "dhw_temperature": {
                 "name": "Brauchwasser"
             },
+            "dhw_target_temperature": {
```
BenPru, line 443: Double entry. See line 144.

custom_components/luxtronik/translations/en.json
```diff
@@ -182,6 +182,24 @@
             },
             "efficiency_pump_minimal": {
                 "name": "Efficiency pump minimal"
+            },
+            "cooling_threshold_temperature": {
+                "name": "T_outside >"
```
BenPru, line 187: Can we find a better name?

custom_components/luxtronik/translations/de.json
```diff
@@ -182,6 +182,24 @@
             },
             "efficiency_pump_minimal": {
                 "name": "Effizienzpumpe Minimal"
+            },
+            "cooling_threshold_temperature": {
+                "name": "T_aussen >"
```
BenPru, line 187: Can we find a better name?

custom_components/luxtronik/translations/en.json
```diff
@@ -422,6 +440,9 @@
             "dhw_temperature": {
                 "name": "Domestic hot water"
             },
+            "dhw_target_temperature": {
```
BenPru, line 443: See L144

BenPru (changes requested):

custom_components/luxtronik/const.py
```diff
@@ -241,6 +241,7 @@ class LuxParameter(StrEnum):
     P0016_HEATING_CIRCUIT2_CURVE_NIGHT_TEMPERATURE: Final = (
         "parameters.ID_Einst_HzMK1ABS_akt"  # 0
     )
+    P0042_MK1_TYPE: Final = "parameters.ID_Einst_MK1Typ_akt"
```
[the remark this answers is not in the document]
rhammen: Sure, I'll change this.

custom_components/luxtronik/climate.py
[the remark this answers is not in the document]
rhammen: Sorry, but I find this very strange. 
Let's say we are in summer, and outside it is 20+ degrees. 
Are you saying that when the heatpump heats the domestic hot water, the heating device is actually (also) heating your underfloor? And/or that it should show heating in that case?
That is not how my system works, and I think it would confuse a lot of users (to see the heating being turned on in summer)...
In my view, the luxtronik device is either heating, or domestic_water, or cooling. One action at a time. 
That is also what the heatpump reports in its status line: There is no heatpump status "heating + domestic water", only "heating" or "hot water".

custom_components/luxtronik/number_entities_predefined.py
rhammen: I'm not sure what your suggesting here, can you explain?

custom_components/luxtronik/translations/en.json
rhammen: Yeah, this may be a bit too cryptic...
"Minimal outdoor temperature for cooling mode" ?

custom_components/luxtronik/translations/de.json
rhammen: "minimale Außentemperatur für den Kühlbetrieb"?

rhammen (commented):
I have resolved all remarks, with the exception of:
   
    LuxOperationMode.domestic_water.value: HVACAction.IDLE.value,
    versus
    LuxOperationMode.domestic_water.value: HVACAction.HEATING.value,

In my view, the desired behavior of the integration is best reflected by using
HVAC_ACTION_MAPPING_HEAT:
    LuxOperationMode.domestic_water.value: HVACAction.IDLE.value,

BenPru:
> In my view, the desired behavior of the integration is best reflected by using HVAC_ACTION_MAPPING_HEAT: LuxOperationMode.domestic_water.value: HVACAction.IDLE.value,

This behavior is okay for me. We can take it.
Just for fun you can have a look at the sensors of the underfloor heating (also in summer). In dhw working mode the underfloor heating gets also heated.

custom_components/luxtronik/climate.py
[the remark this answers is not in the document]
BenPru: The luxtronik firmware reports a lot of time not really the status what it is doing. See for example: [OperationMode correction?](https://github.com/Bouni/python-luxtronik/issues/84)

BenPru (approved):
Thanks for your work and your support.

PR #161, opened 2023-10-18
Improve reported heatpump status

rhammen:
- migrated fixes for thermal desinfection and passive cooling that were already implemented in the old integration to the new integration
- new additional fix for part where thermal desinfection is done using the 2nd heatsource (without heatpump running)


BenPru (approved):
LGTM thank you

PR #167, opened 2023-10-22
fixes and untraslated phrases

qubabos:
fixes and untraslated phrases


BenPru:
@qubabos Can you have a look at the conflicts and resolve them?

Issue #233, opened 2024-01-21
'heating external engery source' is not in the list of options provided


dbu42:
My alpha innotec heat pump has an external energy source temperature sensor "ID_WEB_Temperatur_TEE". If that sensor goes beyond a defined threshold the luxtronik heat pump is blocked and sets status to "heating external engery source". The integration crashes with following exception.

Thanks for checking.

```
Dieser Fehler wurde von einer benutzerdefinierten Integration verursacht

Logger: custom_components.luxtronik2
Source: helpers/update_coordinator.py:300 
Integration: Luxtronik 
First occurred: 17:54:17 (18 occurrences) 
Last logged: 18:11:17

Unexpected error fetching luxtronik2 data: Sensor sensor.luxtronik_0_00_status_line_3 provides state value 'heating external engery source', which is not in the list of options provided
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 300, in _async_refresh
    self.data = await self._async_update_data()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/luxtronik2/coordinator.py", line 98, in _async_update_data
    self.data = await self._async_read_data()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/luxtronik2/coordinator.py", line 102, in _async_read_data
    return await self._async_read_or_write(False, None, None)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/luxtronik2/coordinator.py", line 125, in _async_read_or_write
    self.async_set_updated_data(data)
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 426, in async_set_updated_data
    self.async_update_listeners()
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 182, in async_update_listeners
    update_callback()
  File "/config/custom_components/luxtronik2/sensor.py", line 174, in _handle_coordinator_update
    super()._handle_coordinator_update()
  File "/config/custom_components/luxtronik2/base.py", line 166, in _handle_coordinator_update
    super()._handle_coordinator_update()
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 479, in _handle_coordinator_update
    self.async_write_ha_state()
[...]

PR #234, opened 2024-01-21
Fix for #233

rhammen:
Fix for #233


BenPru:
Thanks. Workaround for [python-luxtronik 0.3.14](https://github.com/Bouni/python-luxtronik/tree/0.3.14)

Issue #248, opened 2024-04-06
time and day related evu events (related #38)


Gifford47:
in my case, the evu is only applicable on monday till friday. but the EVU attributes (especially "time until next evu event") is also count on sunday and saturday.
is it possible to add a day related evu state?
f.e. a new attribute which "holds" the evu days (monday, tuesday, ...). Is  the current day a saturday then "time until next event" counts 2880 minutes for example.
related #38

Issue #38, opened 2022-11-22
Provide a sensor for "next evu event"


BenPru:
A minutes timer number or timestamp with the next evu event.
If the evu is active its the time when the evu ends.
If the evu is not active its the time when the next evu starts.
For example my evu time slots are:
- 10:43-12:46
- 16:42-18:44

Example values:
- At 10:00 the value is 43 minutes.
- At 11:00 the value is 106 minutes.
- At 13:00 the value is 222 minutes.

Gifford47:
in my case, the evu is only applicable on monday till friday. but the EVU attributes (especially "time until next evu event") is also count on sunday and saturday. is it possible to add a day related evu state?

PR #249, opened 2024-04-11
added time and day related evu events (related #38) #248

Gifford47:
i've added a new attribute `SA.EVU_DAYS` and a logic which saves the evu days into the mentioned attribute.
in the `_calc_next_evu_event_minutes(self)` is now a new fraction of code which takes care of the days on which the evu is not present. it calculates the "pause" time and adds it to `EVU_minutes_until_next_event`

Knows Bug:
@BenPru currently i get the following error because of wrong types:
```
  File "/config/custom_components/luxtronik2/sensor.py", line 443, in _wd_txt
    for i in value:
TypeError: 'datetime.time' object is not iterable
```

NOTE: i`ve changed `_attr_cache: dict[SA, object] = {}`

### DIFF
--- a/custom_components/luxtronik/const.py
+++ b/custom_components/luxtronik/const.py
@@ -731,6 +731,7 @@ class SensorAttrKey(StrEnum):
     EVU_SECOND_START_TIME = "EVU_second_start_time"
     EVU_SECOND_END_TIME = "EVU_second_end_time"
     EVU_MINUTES_UNTIL_NEXT_EVENT = "EVU_minutes_until_next_event"
+    EVU_DAYS = "EVU_days"
     TIMESTAMP = "timestamp"
     CODE = "code"
     CAUSE = "cause"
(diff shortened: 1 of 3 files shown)


Gifford47:
@rhammen @ggajoch can you look at this PR please? I´m struggling with the [above ](https://github.com/BenPru/luxtronik/pull/249#issue-2237295727) error.

Gifford47:
@BenPru can you help me please?

BenPru:
@Gifford47 Sorry for the delay.
Please let me publish a new stable release for ha 2024.10.* and after that we can merge your changes.

BenPru (approved):

Gifford47:
The PR is not finalized. As I've wrote in the beginning, I have a issue with the correct datetime type. Can you please help me out?

s-kurz:
@Gifford47 I have the same issue and I'm pretty sure it originates from the attribute restoration as this always datetime. 

```
    def _restore_attr_value(self, value: Any | None) -> Any:
        if value is None or ":" not in str(value):
            return time.min
        vals = str(value).split(":")
        return time(int(vals[0]), int(vals[1]))
```

I did create a fix https://github.com/s-kurz/luxtronik/commit/e42a7899923da5133786647bda661a0dbc724a81 but I'm still testing it. I'm actually not entirely happy with my solution as I feel it would be better to rewrite the _restore_attr_value signature to provide the key as well

Gifford47:
Thx. You can check in line 456 with isinstance () if it's a list, then loop through it's elements.

s-kurz:
@Gifford47 Thx. I'll try that. I thought we had a string at this point

Gifford47:
@s-kurz any progress with the last commit?

s-kurz:
@Gifford47 I did adapt it and for me its running without issue now

PR #264, opened 2024-08-09
Be able to track history of room temperature

ndeunk:
The 'Heat room temperature' has no _state_class_ defined. Therefore it is not tracked in by the recorder into the long term statistics. 

https://developers.home-assistant.io/docs/core/entity/sensor/#long-term-statistics

Added the state_class to the entity.


BenPru (approved):
LGTM

PR #277, opened 2024-10-02
add current power comsumption sensor

s-kurz:
adds the sensor for the current power consumption given by Unknown_Calculation_268

grafik


s-kurz:
I added the minor_88 since its the version used with the energy sensors

PR #283, opened 2024-10-09
restore evu date attribute correctly to solve the TypeError: 'datetime.time'

### LINKED ISSUE #281: Update 2024-10-05-beta multiple sensors disappearing
Hi!

I am currently running version 2024-09-30 with great satisfaction and wanted to thank you for the timely update to the integration.

After getting the notification about 2024-10-05-beta I decided to try that out as well, but this resulted in about half my sensors and all my switches disappearing from the integration. I quickly reverted and everything is fine again, nothing of value lost, just wanted to let you know that - at least for me - there seems to be some problematic change in the latest beta.

s-kurz:
restore the evu dates attribute and solve the TypeError: 'datetime.time' issue 
fixes #249 
fixes #281

### DIFF
--- a/custom_components/luxtronik/sensor.py
+++ b/custom_components/luxtronik/sensor.py
@@ -396,8 +396,10 @@ class LuxtronikStatusSensorEntity(LuxtronikSensorEntity, SensorEntity):
             return None
         evu_hours = (24 if evu_time < time_now else 0) + evu_time.hour
         weekday = datetime.today().weekday()
-        if not isinstance(self._attr_cache[SA.EVU_DAYS], list):
-            self._attr_cache[SA.EVU_DAYS] = list() 
+        if isinstance(self._attr_cache[SA.EVU_DAYS], str):
+            self._attr_cache[SA.EVU_DAYS] = self._attr_cache[SA.EVU_DAYS].split(',')
+        elif not isinstance(self._attr_cache[SA.EVU_DAYS], list):
+            self._attr_cache[SA.EVU_DAYS] = list()
         evu_pause = 0
         if not self._attr_cache[SA.EVU_DAYS] and weekday not in self._attr_cache[SA.EVU_DAYS]:
             evu_pause += (24 - datetime.now().hour)*60 - datetime.now().minute
--- a/custom_components/luxtronik/sensor.py
+++ b/custom_components/luxtronik/sensor.py
@@ -459,7 +459,7 @@ class LuxtronikStatusSensorEntity(LuxtronikSensorEntity, SensorEntity):
                     if day == name: 
                         vals.append(idx)
                         break
-            if len(vals):
+            if vals:
                 return vals 
         return time.min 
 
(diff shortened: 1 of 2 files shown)


Gifford47:
Can this PR already be merged?

s-kurz:
add min firmware version to the current power consumption sensor in alignment with the energy sensors min firmware 
fixes #277

Gifford47:
In my case the attribute EVU_DAYS isn't set correctly. It's always "" (empty string). Does it works on your own?@s-kurz

s-kurz:
@Gifford47 Its similar for me. I just fixed the issue that came from the attribute restoration. To test this part I manually change the attributes and then did a restart, which does restored them correctly. I never checked the logic of how the dates are actually set, nor am I sure if they should be set for my setup

Gifford47:
Ok, thanks for the info!
I'll test it the next days

Gifford47:
@s-kurz can you have a closer look why the attribute "SA.EVU_DAYS" isn't set correctly? I didn't find the bug yet...

Gifford47:
Ok, basic logic works so far. The only problem is that only the current day will be saved into attribute. The other days before are removed when a new day is appended... I cannot find why
@s-kurz @BenPru could you have a look please?

Gifford47:
@BenPru Please don´t merge yet, there's a smaller bug in the logic. I'll fix it in a few days.

custom_components/luxtronik/sensor.py
```diff
@@ -394,16 +396,18 @@ def _calc_next_evu_event_minutes(self) -> int | None:
             return None
         evu_hours = (24 if evu_time < time_now else 0) + evu_time.hour
         weekday = datetime.today().weekday()
+        if not isinstance(self._attr_cache[SA.EVU_DAYS], list):
```
Gifford47, line 399: here's a bug. it has the checked if the attribute is a string. if yes, convert it to a list to append a new weekday. Please replace with the following:

`
        if isinstance(self._attr_cache[SA.EVU_DAYS], str):
            self._attr_cache[SA.EVU_DAYS] = self._attr_cache[SA.EVU_DAYS].split(',')
        elif not isinstance(self._attr_cache[SA.EVU_DAYS], list):
            self._attr_cache[SA.EVU_DAYS] = list()		# failsafe
`

Gifford47 (changes requested):
on line 399 is a bug. it has to be checked if the attribute is a string. if yes, convert it to a list to append a new weekday. Please replace with the following:

``` 
        if isinstance(self._attr_cache[SA.EVU_DAYS], str):
            self._attr_cache[SA.EVU_DAYS] = self._attr_cache[SA.EVU_DAYS].split(',')
        elif not isinstance(self._attr_cache[SA.EVU_DAYS], list):
            self._attr_cache[SA.EVU_DAYS] = list()		# failsafe
```

s-kurz:
@Gifford47 I made the change. But as I said I don't have a good way to test it since EVU is never active for me. I would propose we merge it at this state wether the evu dates are working now or not, so we at least fix the other more problematic issues with the disapperaing/unavailable sensors

PR #286, opened 2024-10-14
Fixed translation errors in Dutch. 

Travelbacon:
E.g. Domestic water was translated as industrial water.  Seems like they were translated via a translate tool?


BenPru (approved):
LGT

PR #298, opened 2024-11-18
fixed EVU_DAYS Attribute to bet set correctly [ref #249]

Gifford47:
it's tested and everything works as expected:
image


Gifford47:
Works now for over a week. Can be merged

Gifford47:
@BenPru could you merge it please?

Issue #280, opened 2024-10-06
Warmwater set target temperature


jumahrauf:
Hello,
I am using 2024.10.5-Beta with luxtronik firmware V3.90.1.

Actually I am not able to set the warmwater target temperature. 
Where can I see in a log what the current problem is?

rhammen:
Settings --> System --> Logs.
Then, using the search-bar at the top of the page, search for "luxtronik"

jumahrauf:
Logger: py.warnings
Quelle: custom_components/luxtronik2/coordinator.py:121
Integration: Luxtronik
Erstmals aufgetreten: 6. Oktober 2024 um 20:03:47 (2 Vorkommnisse)
Zuletzt protokolliert: 10:20:07

/config/custom_components/luxtronik2/coordinator.py:121: RuntimeWarning: coroutine 'DataUpdateCoordinator.async_request_refresh' was never awaited self.async_request_refresh()

blue-bean:
In my case (SWCV92H3) with firmware 3.90.1 the target temperature is 49° and the coverage temperatur of the heat pump is shown as 75° (at the control panel). In HA the target temperature is then shown with 75°.
Maybe the wrong parameter ID is usd?
In the list of [parameters ](https://github.com/Bouni/python-luxtronik/blob/main/luxtronik/parameters.py )there exist 
    2: Celsius("ID_Einst_BWS_akt", True),
and 
   105: Celsius("ID_Soll_BWS_akt", True),

The fixed 75° for the coverage temperature seems th be a bug in the 3.90.1. Usally this number should be the same as the target temperatur.

blue-bean:
> Did you try writing parameter 105: Celsius("ID_Soll_BWS_akt", True), and does that change the hot water temperature on the heatpump?

No, i don't know how to do this.
Did you try to set hot target temperature by 105: Celsius("ID_Soll_BWS_akt", True) using firmware 3.79?

jumahrauf:
seems like with firmware 3.90 the accessability is kind of limited. 
could somebody check i he can set the sensor "luxtronik_331009_0996_dhw_target_temperature"?

blue-bean:
> seems like with firmware 3.90 the accessability is kind of limited. 
Not 3.90.0, but 3.90.1

jumahrauf:
[...]

PR #305, opened 2024-12-16
Using ID_Soll_BWS_akt for hot water target temperature

blue-bean:
Using ID_Soll_BWS_akt for hot water target temperature
See https://github.com/BenPru/luxtronik/issues/280

### DIFF
--- a/custom_components/luxtronik/const.py
+++ b/custom_components/luxtronik/const.py
@@ -272,7 +272,7 @@ class LuxParameter(StrEnum):
 
     UNSET: Final = "UNSET"
     P0001_HEATING_TARGET_CORRECTION: Final = "parameters.ID_Einst_WK_akt"
-    P0002_DHW_TARGET_TEMPERATURE: Final = "parameters.ID_Einst_BWS_akt"
+    P0105_DHW_TARGET_TEMPERATURE: Final = "parameters.ID_Soll_BWS_akt"
     P0003_MODE_HEATING: Final = "parameters.ID_Ba_Hz_akt"
     P0004_MODE_DHW: Final = "parameters.ID_Ba_Bw_akt"
     P0011_HEATING_CIRCUIT_CURVE1_TEMPERATURE: Final = "parameters.ID_Einst_HzHwHKE_akt"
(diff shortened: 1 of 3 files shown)


Alkaramazov (approved):
Works nice

stefanovich1 (approved):

Alkaramazov:
Tested on [2025.1.4-Beta](https://github.com/BenPru/luxtronik/releases/tag/2025.1.4-Beta)
HA Core 2025.1.1
HA Supervisor 2024.12.3

It works correctly

colognebroker:
For me it is not working with V3.90.2-12457 firmware and also. not for 3.90.3

blue-bean:
> For me it is not working with V3.90.2-12457 firmware and also. not for 3.90.3

Which version of HA and which version of this integration? How did you the changes described in my commit?

rhammen:
I refer to my comment on Oct 22, 2024 in #280:

"So with FW3.79, still ID_Einst_BWS_akt needs to be used (as does the 2024.10.5-Beta luxtronik integration)
**This means any change in the integration to use ID_Soll_BWS_akt instead of the currently used ID_Einst_BWS_akt should be firmware version dependant (>= .90)**"

In your commit, I see that you simply change the target parameter to ID_Soll_BWS_akt where it currently is ID_Einst_BWS_akt.
I do not see the target parameter being a firmware dependant?
That means this PR will break the integration for users who are not on FW 3.90.x !

Fixing the problem for some users and at the same time breaking it for others is not the way forward.
So please implement a FW dependant setting, as that apparently is what is needed now since the Luxtronik FW versions require different target parameters.

eliaslecomte:
Could we make the variable based on the running firmware version?
This way it's not broken for users that didn't update their firmware **and** it doesn't break when you update your firmware 🤓?

cpetry:
I'm currently at exactly that problem. I did update firmware and I can't set BW temp...

AJediIAm:
@blue-bean : it doesn't seem to work in combinatie with other changes in the latest developer release. Can you have a look at it?
It might be a mistake on my side. Can you check?

PR #357, opened 2025-05-30
Feature/hot water for firmware 90

### LINKED ISSUE #280: Warmwater set target temperature

siedi:
This should fix the annoying bug for firmwares >= 3.90.1 to not being able to set the hot water temperature any more:

The change introduces a way to compare the firmware for min and max version (the complete version, not only the minor number, as we need to compare against 3.90.1 vs. 3.90.0), but still keeps the old logic for compatibility.

The patch uses  ID_Soll_BWS_akt for hot water target temperature now.
Fixes #280

### DIFF
--- a/custom_components/luxtronik/number_entities_predefined.py
+++ b/custom_components/luxtronik/number_entities_predefined.py
@@ -1,4 +1,7 @@
 """Luxtronik number sensors definitions."""
+
+from packaging.version import Version
+
 # region Imports
 from homeassistant.components.number import NumberMode
 from homeassistant.components.sensor import SensorDeviceClass
@@ -12,6 +15,7 @@ from homeassistant.helpers.entity import EntityCategory
 
 from .const import (
     DeviceKey,
+    FirmwareVersionMinor,
     LuxCalculation as LC,
     LuxParameter as LP,
     LuxVisibility as LV,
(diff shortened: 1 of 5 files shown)


tmp-hallenser:
@BenPru Any chance you can review and provide feedback/approve/merge the PR? I think, many people would appreciate a fix for the hot water issue

custom_components/luxtronik/number_entities_predefined.py
AJediIAm, line 1: I would expect import Version to be underneath # region Imports instead of above.

AJediIAm:
I copied the code to test the change and it works! 
I can adjust the temperatures within the allowed range. 

I'm not an experience python programmer, but the code looks good. It contains some small touch-ups on styling besides the version-based switch to the new parameter. 

There are some unrelated luxtronik2 errors and warnings.
I'll be using this version from tomorrow onward in production and report any issues I encounter.

AJediIAm:
This may be a duplicate of https://github.com/BenPru/luxtronik/pull/305

PR #360, opened 2025-06-20
Update update.py

AJediIAm:
Fix firmware version detection from file for 3.91.0


BenPru:
Thanks for your change. 
Can you extend the code comments with all filename examples?

martinl:
I created new PR https://github.com/BenPru/luxtronik/pull/366

Issue #361, opened 2025-06-20
update.heatpump_firmware sensor fails with V3.91.0


AJediIAm:
The available firmware detection fails due to a change in the filename. 

![Image](https://github.com/user-attachments/assets/b0a6c497-6f6b-445b-84de-6d51255ad145)

https://github.com/BenPru/luxtronik/pull/360 is created to solve this.

Please test the change on older firmware versions and report if it works or not.

martinl:
I added an updated regexp to PR comments to retain current behaviour

AJediIAm:
Fixed by @martinl

PR #366, opened 2025-07-05
Fix firmware version check for V3.91

martinl:
This PR fixes firmware version check for V3.91 version that has new filename format (wp2reg-V3.91.0_d0dc76bb) https://github.com/BenPru/luxtronik/issues/361


BenPru (approved):
LGT

Issue #346, opened 2025-03-08
Add option for RBE functionality


JarnoHilverts:
Hi all,

This is the first issue I’m opening, hopefully everything is described clearly.

In the last few days, I’ve been working to control the RBE that’s installed in the kitchen using this integration. I’ve managed to do this, but I’m not completely satisfied with the code, so the solution still seems a bit messy.

Parameter 1148 is used to select the desired target temperature from the RBE. I’ve tested this using the service to adjust the temperature by modifying parameter 1148. It’s important that the value is sent without a decimal. Therefore, 21.5 degrees is 215.

Service call:
```yaml
action: luxtronik2.write
data:
  parameter: Unknown_Parameter_1148
  value: 215
```

Below is a description of how this has been implemented in the code. Could someone help further develop this to create a pull request, to further implement this functionality into this integration? 
Personally, I don't know how the integration handles it when no RBE is connected to the system. Maybe this can be caught somewhere, or someone else can test this.

Files changed: 
- Const.py
- Climate.py

In the Const.py file the parameter 1148 needs to be added:
```python
P1148_HEATING_TARGET_TEMP_RBE: Final = "parameters.Unknown_Parameter_1148"
```

The climate.py file the `THERMOSTATS` for heating needs to be modified so that the actual desired temperature matches that of the RBE. This was done by commenting out the `luxtronik_key_target_temperature`. Additionally, the code for writing the target_temperature (parameter 1148) has been modified. Because no decimal number is sent, the value is multiplied by 10. The rest of the code is similar to the code in the water_heater.py:
```python
THERMOSTATS: list[LuxtronikClimateDescription] = [
    LuxtronikClimateDescription(
        key=SensorKey.HEATING,
        hvac_modes=[HVACMode.HEAT, HVACMode.OFF],
        hvac_mode_mapping=HVAC_MODE_MAPPING_HEAT,
        hvac_action_mapping=HVAC_ACTION_MAPPING_HEAT,
        preset_modes=[PRESET_NONE, PRESET_AWAY, PRESET_BOOST],
        supported_features=ClimateEntityFeature.PRESET_MODE
        | ClimateEntityFeature.TURN_OFF  # noqa: W503
        | ClimateEntityFeature.TURN_ON  # noqa: W503
        | ClimateEntityFeature.TARGET_TEMPERATURE,  # noqa: W503
        luxtronik_key=LuxParameter.P0003_MODE_HEATING,
[...]

PR #371, opened 2025-07-07
Added option for RBE functionality

JarnoHilverts:
A pull request for issue #346

### DIFF
--- a/custom_components/luxtronik/const.py
+++ b/custom_components/luxtronik/const.py
@@ -394,6 +394,7 @@ class LuxParameter(StrEnum):
     # ? P1138_SWIMMING_POOL_ENERGY_INPUT: Final = "parameters.Unknown_Parameter_1138" -->
     P1139_COOLING_ENERGY_INPUT: Final = "parameters.Unknown_Parameter_1139"
     P1140_SECOND_HEAT_GENERATOR_AMOUNT_COUNTER: Final = "parameters.Unknown_Parameter_1140"
+    P1148_HEATING_TARGET_TEMP_RBE: Final = "parameters.Unknown_Parameter_1148"
 
 
 # endregion Lux parameters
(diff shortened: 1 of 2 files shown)


rhammen (commented):
as far as I can see, def async_set_temperature() is also used by the cooling climate thermostat. Your current changes do not look at which luxtronik_key_target_temperature is defined, but simply always write to LuxParameter.P1148_HEATING_TARGET_TEMP_RBE
So the current changes will break the cooling thermostat and the generic way things are setup.

rhammen:
Also, the existance of LuxParameter.P1148_HEATING_TARGET_TEMP_RBE depends on the firmware version of the heatpump.
So we must check on that and deal with that, because otherwise it will break the integration for people with older firmware.

rhammen:
I think the above remarks must be addressed and verified before merging this PR.

custom_components/luxtronik/const.py
```diff
@@ -394,6 +394,7 @@ class LuxParameter(StrEnum):
     # ? P1138_SWIMMING_POOL_ENERGY_INPUT: Final = "parameters.Unknown_Parameter_1138" -->
     P1139_COOLING_ENERGY_INPUT: Final = "parameters.Unknown_Parameter_1139"
     P1140_SECOND_HEAT_GENERATOR_AMOUNT_COUNTER: Final = "parameters.Unknown_Parameter_1140"
+    P1148_HEATING_TARGET_TEMP_RBE: Final = "parameters.Unknown_Parameter_1148"
```
BenPru, line 397: RBE ist german for "Raum Bedien Einheit". Can we call it "ROOM_THERMOSTAT"?
--> P1148_HEATING_TARGET_TEMP_ROOM_THERMOSTAT

AJediIAm (approved):
Works for me

rhammen:
No, #400 is a fix for another regression issue.

Issue #384, opened 2025-07-21
Integration broken after update to 2025.07.20-beta


rhammen:
2025-07-21 21:53:51.690 INFO (Dummy-1) [homeassistant.setup] Setting up luxtronik2
2025-07-21 21:53:51.690 INFO (Dummy-1) [homeassistant.setup] Setup of domain luxtronik2 took 0.00 seconds
2025-07-21 21:53:51.690 ERROR (Dummy-1) [homeassistant.config_entries] Error importing platform config_flow from integration luxtronik2 to set up luxtronik2 configuration entry: Platform luxtronik2.config_flow not found

No luxtronik2 entities available in HA.

AJediIAm:
I'm using the latest version on firmware 3.90 without issue.
The current main branch might have an issue, but 2025.7.20 works fine for me.

Do you have the same issue with 2025.7.19?
I was thinking that we might need more debug logging for troubleshooting purposes.

rhammen:
I tried 2025.7.19 --> same issue.
Then I re-downloaded 2025.7.20 again, AND I updated HA to 2025.7.2. 
Updating HA to 2025.7.2 solved the problem!
Before I was on HA 2025.1.3.
So I think you may need to update the manifest.json file, which currently mentions  "homeassistant": "2024.8.0",

AJediIAm:
I updates the manifest to 2025.6 https://github.com/BenPru/luxtronik/pull/385. I don't recall any major changes in 2025.7 so this should provide a bit more flexibility. Not very scientific to just guess a compatible version, so I hope we'll get some more feedback with HA versions which do or do not work.

PR #385, opened 2025-07-22
Increase minimal HA version manifest.json

AJediIAm:
The latest version of the release doesn't seem to work with older versions of home assistant. It's unclear from which version it breaks.


AJediIAm:
Fixes https://github.com/BenPru/luxtronik/issues/384

Issue #390, opened 2025-07-25
Workflow errors: manifest, translations, entity sensor definition


BenPru:
# Workflow errors: manifest, translations, entity sensor definition

This issue summarizes the errors detected in the workflow run [Validate with hassfest](https://github.com/BenPru/luxtronik/actions/runs/16532309572/job/46760230180) at commit `58c49d8bd4ff4a51849de25f30a94853d2442fb9`.

---

### Manifest (`custom_components/luxtronik/manifest.json`)
- Extra key not allowed: `homeassistant` (value: `'2025.6.0'`)
- Domain does not match directory name

### Translations (`custom_components/luxtronik/translations/en.json`)
- Invalid translation key `-1`: Keys must be `[a-z0-9-_]+` and cannot start/end with hyphen/underscore
- Invalid translation key `heatpump running`
- Invalid translation key `hot water`
- Invalid translation key `pump forerun`

### Entity sensor definition (likely in `custom_components/luxtronik/`)
- Extra key not allowed: `error_reason.cause`
- Extra key not allowed: `error_reason.remedy`

---

**Reference:**  
- [Workflow run](https://github.com/BenPru/luxtron)


AJediIAm:
Regarding: Domain does not match directory name

There is no perfect solution, so the simplest solution should be preferred. The simple solution would be to change the domain back to Luxtronik.

Are there arguments why changing the directory would be better than changing the domain?

AJediIAm:
> > Regarding: Domain does not match directory name
> > There is no perfect solution, so the simplest solution should be preferred. The simple solution would be to change the domain back to Luxtronik.
> > Are there arguments why changing the directory would be better than changing the domain?
> 
> All Sensors are renamed after that. Complete Sensor History ist gone and everyone have to Change The Sensor names in all automations, Scripts, Dashboard, ... I prefer changing the directory name.

I assumed existing installations wouldn't change and that a new directory would force a new instance to be created. 
Giving it a second thought, I simply don't know. I'll test what happens.

AJediIAm:
As it turns out, renaming the domain doesn't work. 
The translation function of home assistant seems to have an internal reference to Luxtronik2 so it fails. This may occur at other places as well. There is a change I didn't correctly update all references within the integration which caused the error. I can't rule this out.

[...]

PR #393, opened 2025-08-05
hassfix

AJediIAm:
Move min ha version


AJediIAm:
Partial fix for https://github.com/BenPru/luxtronik/issues/390

PR #399, opened 2025-08-16
Rename github folder

AJediIAm:
Rename the GitHub folder to `custom_components/luxtronik2`. This has not effect on new and existing installations since HACS already uses the domain name as the folder name.


AJediIAm:
Hold-on, still need to add the files back...

AJediIAm:
It's not a good idea to do this now since any open activities will need to be redone. Parking this for later.

PR #401, opened 2025-08-24
Attempt to fix the RBE feature

rhammen:
This fixes the writing of the target temperatures for old firmware.
And it *should* also keep the RBE feature working.
Please test that part, since my firmware doesn't have the P1148_HEATING_TARGET_TEMP_ROOM_THERMOSTAT parameter, so I cannot test that part.

### DIFF
--- a/custom_components/luxtronik/climate.py
+++ b/custom_components/luxtronik/climate.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 from dataclasses import asdict, dataclass
 from typing import Any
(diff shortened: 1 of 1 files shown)


AJediIAm:
I have tested this change a few times and it doesn't produce any errors. I'd like to do more functionality testing in combination with the other changes in a new developer release, so I'm merging the PR knowing that there could be a few bugs in it.

AJediIAm (approved):

PR #402, opened 2025-08-26
Startup lockup

AJediIAm:
Set timeout earlier and add exception handling for timeout.


AJediIAm:
@BenPru : Can you have a look at this?
It works, but I'm not familiar enough with this part of the code to trust my own judgement.

BenPru (approved):
LGT

PR #404, opened 2025-08-26
Resolve Status-Cooling flickering

rhammen:
When cooling, HP status was displayed as idle for 0.001s every 1-2 minutes. Changed the detection and moved it from sensor.py to common.py


AJediIAm:
I tested the change and it breaks the cooling status. This needs to be fixed before a new release can be created.

rhammen:
So Status = Idle (no request), and Status 3 = heating, and then the Heatpump cooling...
Correct?
Interesting :-)
Is this Active cooling?

PR #410, opened 2025-08-30
Changed lux_helper.py to fix discover logic #312

rhammen:


rhammen:
> Do I assume correctly that the info logging is temporary and that we'll change it back to debug once it is fixed?

Yeah, I really think we need more logging to able to figure out the various issues people reported when adding the integration.
OF course, if these installation issues are resolved, we can go back to debug.

AJediIAm:
Let's add the logging in the beta release and clean it up in in the final releases. 
If we add a ToDo comment to it, we can easily find it back.

PR #413, opened 2025-09-01
Translation and code fixes for all hassfest errors, except "Error: R] [MANIFEST] Domain does not match dir name"

rhammen:
Except for the error
"Error: R] [MANIFEST] Domain does not match dir name"
all other hassfest checks now pass. So it fixes most of #390 

Link to successfull hassfest run with domain temporarily set to 'luxtronik' to match dir name:
https://github.com/rhammen/BenPru_luxtronik/actions/runs/17386929553/job/49354846650

Please test/check translations for heatpump states show correctly in HA for non-english setup before merging.


AJediIAm (approved):
Too much to review every line and change thoroughly. Since it's a low impact change, it's good to go for the next development release.

AJediIAm:
My test setup is English, so I haven't tested this individual change with a non-english setup yet. I'll mention it in the release notes.

Issue #116, opened 2023-09-02
RuntimeWarning: coroutine 'LuxtronikThermostat._async_control_heating


Crashman1983:
Don't know what that means in practice....

Logger: py.warnings
Source: custom_components/luxtronik2/climate.py:254 
Integration: Luxtronik 
First occurred: 07:54:14 (1 occurrences) 
Last logged: 07:54:14

/config/custom_components/luxtronik2/climate.py:254: RuntimeWarning: coroutine 'LuxtronikThermostat._async_control_heating' was never awaited self._async_control_heating()

BenPru:
See the issue #75.

Issue #108, opened 2023-07-10
coroutine 'LuxtronikThermostat._async_control_heating' was never awaited self._async_control_heating()


augi:
This error originated from a custom integration.

Logger: py.warnings
Source: custom_components/luxtronik2/climate.py:254
Integration: Luxtronik
First occurred: June 30, 2023 at 7:02:03 PM (9 occurrences)
Last logged: July 9, 2023 at 9:29:18 PM

/config/custom_components/luxtronik2/climate.py:254: RuntimeWarning: coroutine 'LuxtronikThermostat._async_control_heating' was never awaited self._async_control_heating()

Maybe there should be something like `asyncio.get_event_loop().run_in_executor()` used. I'm using `2023.04.30`.

BenPru:
See the branch https://github.com/BenPru/luxtronik/tree/core-preview
I have redesigned the hole integration in this branch to publish it as core ha integration. But I have no support and atm not enough time to finalize the work on it.
See the issue #75.

BenPru:
Unfortunately there is no new pypi python-luxtronik version yet.
In order to get and display the newly calculated values ​​immediately after a write process, a sleep of 1000ms is good. <800ms did not always deliver the new data.
However, sleep is prohibited in the HA core. An async would exit the thread and delay or make the GUI update impossible.
As a workaround, I'm simply considering closing the socket connection between write and read and re-establishing it. I guess this can make sleep unnecessary and we can still get the new data.

AJediIAm:
Has this issue been resolved?

PR #432, opened 2025-09-27
Fix coroutine never awaited

rhammen:
The changes in the PR fix the long-time issues with this integration regarding ASYNC behavior. Example issues are in #108 #116 
The changes to the code are major, so this should end up in a separate beta-release, focussing on just this PR.
There are also some changes to the functionality that should be discussed if they are acceptable, or that additional changes are needed.

### DIFF
--- a/custom_components/luxtronik/__init__.py
+++ b/custom_components/luxtronik/__init__.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import CONF_TIMEOUT, Platform as P
 from homeassistant.core import HomeAssistant
+from homeassistant.exceptions import ConfigEntryNotReady
 from homeassistant.helpers import device_registry as dr
 from homeassistant.helpers.entity_registry import (
     async_get,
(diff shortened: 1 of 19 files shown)


rhammen:
@AJediIAm @BenPru The new code lets HA handle the data updates.
A consequence of this is that in the new code - at this moment - you can only specify 1 UPDATE_INTERVAL. I choose UPDATE_INTERVAL_NORMAL for now.
Question is: is it acceptable to use only 1 UPDATE_INTERVAL, or is it a major issue/drawback?

Main reason for the change is that the update logic can be much simpler.
And since we read the data for all parameters and calculations from the heatpump always in 1 action, at the same time: is there a real need to have different update frequencies?

One need in the past was that when writing to the heatpump, we wanted the data to be read-back and updated in HA immediately. I think I have solved that in this PR, it works OK with my heatpump. --> Please try it, if you can.

@BenPru In the old code I see we switch to UPDATE_INTERVAL_FAST if  bool(self.get_value(LC.C0044_COMPRESSOR))
Why do we need/do that, and do we still need it?
If I understand why we would still need it, I can have a look on how that could be done with the new code.

Then there are some entities with slower updates frequencies. Is it a problem to put those at NORMAL?

rhammen:
> @rhammen: can you have a quick look at the merge conflicts? I have some slight doubts about what to keep.

Should be ok now, but let's test :-)

AJediIAm:
I want to test this change for a couple of days, so I'll create a new beta release first and leave this PR for the next beta release which will follow shortly.

PR #481, opened 2025-11-11
Added thermal power limit

Codectory:
Added thermal power limit from firmware V3.92.0. I forgot if and what I have to do with the visibilities. It's working fine for me with V3.92.1, so maybe someone can check, if it is necessary to change the visibility entries.


AJediIAm:
I'm trying to find out how these settings work exactly. 
Do I understand correctly the new settings are listen on page 32 of the [manual](https://mw.ait-group.net/files/docs/EN/A0220/83055400.pdf)?
- control heating
- output heat max
- output dhw max
- output cooling

PR #490, opened 2025-11-23
Change factor value from 0.1 to 0.01 for additional_heat_generator_amount_counter 

GermanFab:
The factor was incorrect for the auxiliary heating.
Instead of 54.9 kWh, 549 kWh was displayed.
With this changes it works with my:
- **Alpha-Innotec:** LWD 70A
- **Software Version:** V2.88.3
- **Heatpump Typ:** LD7

### DIFF
--- a/custom_components/luxtronik/sensor_entities_predefined.py
+++ b/custom_components/luxtronik/sensor_entities_predefined.py
@@ -350,7 +350,7 @@ SENSORS: list[descr] = [
         native_unit_of_measurement=UnitOfEnergy.KILO_WATT_HOUR,
         invisible_if_value=0.0,
         visibility=LV.V0324_ADDITIONAL_HEAT_GENERATOR_AMOUNT_COUNTER,
-        factor=0.1,
+        factor=0.01,
         native_precision=1,
         update_interval=UPDATE_INTERVAL_VERY_SLOW,
     ),


dummabua:
I skipped this release and went directly to 2026.01.27 yesterday. It seems that this factor change was not necessary for my Novelan Helox 11 (Type: LP11V, Luxtronik SW: V3.92.1).

Before the update my HA and the Luxtronik WebGUI showed a total of 18h aux-heater-usage consuming 157kWh, which sounds plausible for a 9kW aux-heater. With this update it changed from 157kWh to 15,7kWh.

GermanFab:
I activated the heating rod today to check the values. Strangely enough, the values do not match today. Perhaps this is due to the 5-minute aggregation. However, they correlate in terms of the increase.
How can you check which attributes are the right ones?
Is it possible to set the factor as a parameter?

Luxtronik Key 1059 parameters.ID_Waermemenge_ZWE
sensor.luxtronik_291029_0578_additional_heat_generator_amount_counter

image

dummabua:
Hi,

has there been any update or investigation on this? Version 2026.01.27 still shows the value with factor 10 too small on my Novelan Helox 11 with most recent SW 3.92.2.

GermanFab:
Hello everyone,


Unfortunately, the old AIT was defective. It was replaced with a new AIT Hybrox.
Kudos to the integration team—the new heater was automatically detected and integrated!
However, it seems there is no longer a kWh meter for the auxiliary heater, so I can no longer validate it with the new version.
There is now only an hourly meter, but the factor matches.

grafik

**sensor.luxtronik_...._additional_heat_generator_operation_hours:**
grafik

grafik

grafik

dummabua:
Hi,

same here with the web GUI on my Helox 11, which seems to be essentially the same unit, as both come from AIT with only slightly different housings. It only shows the operational hours of the aux heater, not kWh. BUT: I think this has also been the case with older firmware versions – the web GUI has never shown the consumed kWh, although the value is actually available through the integration. Maybe you just have to activate it in HA, as I remember it being disabled by default.

PS: I changed the value back to 0.1 in the corresponding Python file as a workaround while waiting for this to be fixed in the integration itself 🙂

Kars-de-Jong:
@AJediIAm You may not see this activity on a closed PR, but could you please have a look? I can also report this as an issue if you prefer.

PR #495, opened 2025-12-10
feat: Enum-based DHW mode select entity for ID_Ba_Bw_akt

age-77:


AJediIAm:
@age-77 : could you describe the change for the documentation and release notes? Maybe with a screenshot so I can update the documentation?

age-77:
This change adds a new select entity for the domestic hot water mode based on the existing Luxtronik parameter ID_Ba_Bw_akt (LP.P0004_MODE_DHW).​

Instead of exposing domestic hot water only as a binary on/off switch, Home Assistant now gets an enum-like selector with the full set of operating modes (Off, Automatic, Second heatsource, Party, Holidays).​

The entity both reads the current mode from the controller and allows changing it from HA, so the selected option in Home Assistant always reflects the actual mode on the Luxtronik controller.

image

PR #502, opened 2025-12-25
Fix DHW minimum temperature to match myUplink (30°C)

### LINKED ISSUE #498: Range DHW
I notice that the range of my DHW entity is limited from 40 - 58.
Is this based on a range coming from the device or hardcoded?

In myUplink I can set from 30 to 58, which I actually do. Keep it low at 35 and then increase it to 53 when it's sunny outside.

eliaslecomte:
- Lower the minimum DHW temperature from 40°C to 30°C to match the myUplink interface
- Add `DEFAULT_DHW_MIN_TEMPERATURE` constant for consistency across entities
- Apply the new minimum to both the water heater entity and the DHW target temperature number entities

## Background

The previous minimum of 40°C for the water heater was arbitrarily set during initial implementation. The number entities used 5°C which was also arbitrary. The myUplink interface allows 30-58°C, so 30°C is now used as the authoritative minimum.

No Luxtronik register exists for minimum DHW temperature (only `P0973_MAX_DHW_TEMPERATURE` for maximum), so a hardcoded value is appropriate here 😢.

## Test plan

- [x] Verify water heater entity allows setting temperature down to 30°C
- [x] Verify DHW target temperature number entity allows values down to 30°C
- [x] Confirm heat pump accepts the lower temperature setpoint

## Issues

Fixes #498

### DIFF
--- a/custom_components/luxtronik/const.py
+++ b/custom_components/luxtronik/const.py
@@ -42,6 +42,8 @@ UPDATE_INTERVAL_VERY_SLOW: Final = timedelta(minutes=5)
 
 
 SECOND_TO_HOUR_FACTOR: Final = 1 / 3600
+
+DEFAULT_DHW_MIN_TEMPERATURE: Final = 30.0
 # endregion Constants Main
 
 # region Conf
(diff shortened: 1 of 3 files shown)


eliaslecomte:
Thanks for that added context & information! That should ind not be part of this PR.

AJediIAm:
Thanks for the PR

Issue #500, opened 2025-12-19
Enhancement request: Smart Grid


mhuiskes:
In a recent update of the integration an input has been added `Smart Grid` which is an on/off toggle. When Smart Grid is enabled, it has four operating modes. 

References: (Bundesverband Wärmepumpe, BWP) [https://www.waermepumpe.de/fileadmin/user_upload/bwp_service/SG_ready/SG_Ready_Schnittstelle_1.1.pdf](url)

In Luxtronik these are mapped to EVU1 and EVU2
[https://mw.ait-group.net/files/docs/NL/A0340/83055600.pdf](url)

| EVU 1 | EVU 2 | Bedrijfstoestand |
| AAN (1) |  UIT (O) 1 | (= EVU-blokkering)
| UIT (O) | UIT (O) | 2 (= Verlaagde werking)
| UIT (O) |  AAN (1)| 3 (= Normaal bedrijf)
| AAN (1) | AAN (1) 4 (= Verhoogde werking)

The enhancement request is to provide a status entity (sensor) which mode indicates the smart grid operating status, when smart grid is enabled.

AJediIAm:
> I have EVU1 and EVU2 connected to a shelly relais indeed. Warning: if you work directly on the A1 board: this works with 230V AC on the inputs. I got my installer to place the cable on the board

Do you have a wiring diagram I can add to the documentation? I expect this is a very interesting feature for many users and I'm working on better user documentation.

siedi:
> > I have EVU1 and EVU2 connected to a shelly relais indeed. Warning: if you work directly on the A1 board: this works with 230V AC on the inputs. I got my installer to place the cable on the board
> 
> Do you have a wiring diagram I can add to the documentation? I expect this is a very interesting feature for many users and I'm working on better user documentation.

This is officially documented in the Alpha innotec installation manual. Beyond that, I would not recommend documenting it here. As @mhuiskes has already mentioned, this deals with 230 VAC. If someone wants to do it themselves, they will find the documentation for it. As I said, I would not document it here for liability reasons. In addition, there are different boards. And if the EVU lock has already been installed by the electricity companies, this must also be considered separately.

AJediIAm:
Liability is not an issue. We just share information. You will find loads of wiring schematics on GitHub and the home assistant community forum.

rhammen:
@mhuiskes The smartgrid-evu2-sensors enhancement by @siedi  is part of the latest release, 2026.01.27
Did you test it?
[...]

PR #503, opened 2026-01-01
feat: #500 Add EVU2 and SmartGrid status sensors

siedi:
- Add EVU2 binary sensor (calculation ID 185: ID_WEB_HZIO_EVU2)
  - Add SmartGrid status sensor with 4 states based on EVU+EVU2
  - SmartGrid status: locked, reduced, normal, increased operation
  - Dynamic state-based icons for SmartGrid status
  - Sensor becomes unavailable when SmartGrid switch is disabled
  - Add translations for DE, EN, CS, NL, PL

### DIFF
--- a/custom_components/luxtronik/sensor.py
+++ b/custom_components/luxtronik/sensor.py
@@ -225,9 +225,10 @@ class LuxtronikStatusSensorEntity(LuxtronikSensorEntity, SensorEntity):
         """Handle updated data from the coordinator."""
         if self.entity_description.key == SensorKey.SMART_GRID_STATUS:
             self._update_smart_grid_status()
-        else:
-            # For normal status sensors, use the parent's update logic
-            super()._handle_coordinator_update(data)
+            return
+
+        # For normal status sensors, use the parent's update logic
+        super()._handle_coordinator_update(data)
 
         self._evu_tracker.update(self._attr_native_value)
 


rhammen (changes requested):
and (
                    description.luxtronik_key == LC.UNSET
                    or key_exists(coordinator.data, description.luxtronik_key)
                )

Not all firmware versions may have the LP.P1030_SMART_GRID_SWITCH parameter that you use in the code later. So you need check here that this key exists, otherwise we will get complainst from users using firmware versions that do not yet support this parameter.

Ideally you also check for existence of the keys
LC.C0031_EVU_UNLOCKED and LC.C0185_EVU2
but the risk is significantly less because if LP.P1030_SMART_GRID_SWITCH  exists then these should also exist.

rhammen (commented):
Would it be possible to put the whole code block starting with
`if self.entity_description.key == SensorKey.SMART_GRID_STATUS:`
in a separate function?
So we get
```
       if self.entity_description.key == SensorKey.SMART_GRID_STATUS
            some_function_doing_all_the required_things
       else:
            # For normal status sensors, use the parent's update logic
            super()._handle_coordinator_update(data)
```
This would keep the _handle_coordinator_update() function more readable.

siedi:
> Cool work! Just curious, would it be possible to change the smartgrid status from home assistant? Or are physical relais always required?

Nope, you need the physical relay. There is now way to set EVU/2

PR #540, opened 2026-04-28
fix #539: allow IP change via options dialog and auto-discovery

### LINKED ISSUE #539: IP address change not possible
The IP address of my IP changed due to an internal reorganization of my VLANs.

The integration currently doesn't allow the manual change of the ip address. When I open the option dialog, it first tries to reach the HP, which times out of course and then the dialog just throws an untranslated error.

siedi:
Fixes: #539 

- Options flow now shows the form immediately without connecting first, so users can correct the IP when the heat pump is unreachable. Connection errors appear inline in the form instead of aborting.
- Host and port fields added to the options form.
- DHCP discovery now updates the IP of an existing entry automatically using _abort_if_unique_id_configured(updates=...) instead of just aborting when the unique ID is already known.

### DIFF
--- a/custom_components/luxtronik/config_flow.py
+++ b/custom_components/luxtronik/config_flow.py
@@ -19,6 +19,7 @@ from .const import (
     CONF_HA_SENSOR_PREFIX,
     CONF_MAX_DATA_LENGTH,
     CONFIG_ENTRY_VERSION,
+    DEFAULT_HOST,
     DEFAULT_MAX_DATA_LENGTH,
     DEFAULT_PORT,
     DEFAULT_TIMEOUT,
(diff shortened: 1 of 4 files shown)


AJediIAm:
Nice. 
I'm on holiday, so I can't test it. Do you think it's ready for a beta release?

siedi:
> Nice. I'm on holiday, so I can't test it. Do you think it's ready for a beta release?

Frankly, I would feel better if it is tested beforehand. Though it is just a change in the options dialog and shouldn't affect the normal behavior....you never know.

AJediIAm:
I'll take my time to do the tests. Feel free to poke me next week for feedback

PR #551, opened 2026-05-08
refactor: ♻️ simplify code patterns

svasek:
♻️ **What:** Apply code simplification fixes from ruff's `SIM` and `RUF015` rules.

📋 **Changes:**
- `SIM103`: Replace `if cond: return True; return False` → `return cond`
- `SIM401`: Replace verbose `if key in dict: ... else: ...` → `dict.get(key, default)`
- `SIM910`: Replace `dict.get(key, None)` → `dict.get(key)`
- `RUF015`: Replace `list(...)[0]` → `next(iter(...))`
- `SIM110`: Replace `for` loop + `if` + `return True` → `return any(...)`

💡 **Why:** Reduces code complexity and follows Python idioms. All changes are semantically equivalent — no behaviour change.

📊 **Scope:** 6 files changed (13 insertions, 23 deletions — net -10 lines)

🔗 **Merge order:** 5/7 — merge after #550.

### DIFF
--- a/custom_components/luxtronik/__init__.py
+++ b/custom_components/luxtronik/__init__.py
@@ -2,6 +2,7 @@
 
 # region Imports
 from __future__ import annotations
+
 from typing import Any
 
 from homeassistant.config_entries import ConfigEntry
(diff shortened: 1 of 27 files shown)


AJediIAm:
Maybe some previous ruff improvements have been reverted. We'll see if something is left when everything is done.

PR #573, opened 2026-05-13
fix: 🩹 reduce missing Luxtronik key log level from warning to debug

svasek:
fix: 🩹 reduce missing Luxtronik key log level from warning to debug

🩹 **What:** Change "Not present in Luxtronik data, skipping" log level from `warning` to `debug`.

📋 **Changes:**
- Downgrade log level in 7 platform files: `binary_sensor`, `climate`, `date`, `number`, `sensor`, `switch`, `water_heater`
- Add explanatory comment at each occurrence

💡 **Why:** Not all heat pump models/firmware versions support every parameter or calculation. Missing keys (e.g. `P1158`, `P1175–P1178`, `C0268`) are expected on many setups and not an error condition. Logging them as `warning` causes unnecessary log spam on every integration setup.

### DIFF
--- a/custom_components/luxtronik/binary_sensor.py
+++ b/custom_components/luxtronik/binary_sensor.py
@@ -39,7 +39,9 @@ async def async_setup_entry(
         if not key_exists(coordinator.data, i.luxtronik_key)
     ]
     if unavailable_keys:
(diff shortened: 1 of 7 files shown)


AJediIAm:
I think changing the logging to debug is a good change. @rhammen : any objections?

PR #576, opened 2026-05-13
fix: 🩹 import EntityCategory from homeassistant.const

svasek:
# Import `EntityCategory` from `homeassistant.const`

Part of #583. Resolves §T5 from the Pylance type-check review.

## Problem

6 files import `EntityCategory` from the deprecated path:

```python
from homeassistant.helpers.entity import EntityCategory  # ❌ deprecated
```

`EntityCategory` was relocated to `homeassistant.const` in HA 2023.x. The old path still works at runtime but is flagged by Pyright as `reportPrivateImportUsage`.

## Fix

```python
from homeassistant.const import EntityCategory  # ✅ canonical path
```

## Files changed

| File | Change |
|---|---|
| `date_entities_predefined.py` | Replaced import |
| `number_entities_predefined.py` | Merged into existing `homeassistant.const` import |
| `sensor_entities_predefined.py` | Merged into existing `homeassistant.const` import |
| `switch_entities_predefined.py` | Replaced import |
| `select.py` | Replaced import |
| `update.py` | Merged into existing `homeassistant.const` import |

### DIFF
--- a/custom_components/luxtronik/date_entities_predefined.py
+++ b/custom_components/luxtronik/date_entities_predefined.py
@@ -1,4 +1,4 @@
-from homeassistant.helpers.entity import EntityCategory
+from homeassistant.const import EntityCategory
 
 from .const import (
     DeviceKey,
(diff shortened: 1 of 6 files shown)


AJediIAm:
Thanks 🙏

PR #579, opened 2026-05-13
fix(base): 🐛 align _handle_coordinator_update base signature with overrides

svasek:
# Align `_handle_coordinator_update` base signature with overrides

Part of #583. Resolves §T4 from the Pylance type-check review.

## Problem

The base class `LuxtronikEntity._handle_coordinator_update` is defined with:

```python
def _handle_coordinator_update(self, force: bool = False) -> None:
```

But:
- **All 12 subclass overrides** use `data: LuxtronikCoordinatorData | None = None`
- The base class **itself** calls it with `self.coordinator.data` (a `LuxtronikCoordinatorData`) at line 124
- The `force` parameter is **never used** in the method body

This causes Pyright `reportIncompatibleMethodOverride` in every entity subclass.

## Fix

Change the base class signature to match the actual usage:

```python
def _handle_coordinator_update(self, data: LuxtronikCoordinatorData | None = None) -> None:
```

Added `LuxtronikCoordinatorData` to the imports from `.model`.

## Files changed

- `base.py` — 2 lines: import + signature fix

### DIFF
--- a/custom_components/luxtronik/base.py
+++ b/custom_components/luxtronik/base.py
@@ -28,7 +28,11 @@ from .const import (
     SensorAttrKey as SA,
 )
 from .coordinator import LuxtronikCoordinator
-from .model import LuxtronikEntityAttributeDescription, LuxtronikEntityDescription
+from .model import (
+    LuxtronikCoordinatorData,
+    LuxtronikEntityAttributeDescription,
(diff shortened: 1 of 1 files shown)


AJediIAm:
I don't fully understand this one. @rhammen : can you have a look?

PR #598, opened 2026-05-15
refactor: ♻️ rename integration folder to match domain

svasek:
# Rename integration folder from `luxtronik` to `luxtronik2`

## Motivation

This integration has always used the domain `luxtronik2` (declared in `manifest.json` and `const.py`), but the source code has historically lived in `custom_components/luxtronik/`. This mismatch exists because the project is originally a fork of another Luxtronik integration.

While Home Assistant tolerates this discrepancy for custom components, it causes several problems:

- **Non-standard layout** — HA convention (and requirement for core integrations) is that the folder name matches the domain. Future HA versions may enforce this for custom components as well.
- **Test framework incompatibility** — `pytest-homeassistant-custom-component` expects the folder name to match the domain when using `hass.config_entries.async_setup()`, which prevents writing full integration-level tests.
- **Developer confusion** — contributors need to be aware of this quirk when working with imports, CI paths, and debugging.

### Folder rename

All files moved from `custom_components/luxtronik/` to `custom_components/luxtronik2/` using `git mv` to preserve full file history. No changes to file contents.

### CI & config updates

- `.github/workflows/linter.yml` — updated `ruff` and `codespell` paths
- `.github/workflows/test.yaml` — updated `--cov` path
- `pyproject.toml` — updated `codespell` skip path
- `tests/test_update.py` — updated import path

## What is NOT affected

- **Entity IDs** — all entity IDs use the domain (`luxtronik2`), not the folder name. No change for users.
- **Config entries** — stored under the `luxtronik2` domain. No migration needed.
- **Services** — `luxtronik2.write` remains unchanged.
- **Automations & dashboards** — everything references the domain, not the folder. Zero impact.
- **HACS** — HACS installs integrations based on the `domain` field from `manifest.json`, not the folder name in the repository ([source](https://github.com/hacs/integration/blob/main/custom_components/hacs/repositories/integration.py)). All HACS users already have the integration in `custom_components/luxtronik2/`. This rename simply aligns the repo with what HACS already does.

## Upgrade notes

- **HACS users**: No action needed. HACS already installs to `custom_components/luxtronik2/` based on the manifest domain.
- **Manual installations**: Rename `custom_components/luxtronik/` to `custom_components/luxtronik2/` (or delete and re-copy).


AJediIAm:
Any idea how safe this change is? 
Could this have impact on installed instances?

svasek:
> Any idea how safe this change is? Could this have impact on installed instances?

Hi, 

All entity IDs use the domain, which is already luxtronik2, not the folder name.

This PR contains an automatic migration during the first start, as described above.
There is a new method `_cleanup_old_integration_folder`, in the `__init__.py` file on line 43 (called on line 98).

I believe, it should be safe.

svasek:
Rebased to the latest `main` to resolve conflicts.

AJediIAm:
A good number of user use the original Luxtronik integration and this newer Luxtronik2 integration next to each other. The Luxtronik folder should not be touched. I expect all users of this integration are already in the custom_components/luxtronik2/ folder. 

We could describe it as a potential breaking change or log an error at startup telling the user to reinstall the integration.

svasek:
> A good number of user use the original Luxtronik integration and this newer Luxtronik2 integration next to each other. The Luxtronik folder should not be touched. I expect all users of this integration are already in the custom_components/luxtronik2/ folder.
> 
> We could describe it as a potential breaking change or log an error at startup telling the user to reinstall the integration.

Hi,

How can they use both if they use the same folder name?

Anyway, the automatic deletion of the `luxtronik` folder is applied only if `luxtronik/manifest.json` contains `"domain": "luxtronik2"`.  

So this should be OK if the original integration uses a different domain than `luxtronik2`. 
Original `Bouni/luxtronik` use DOMAIN `luxtronik` BTW.

AJediIAm:
They don't have the same folder name. This integration lives in the Luxtronik2 folder already. Moving installed instances shouldn't be nessesery. Checking the manifest is a smart check.

I haven't done the review yet since I first wanted to discuss the approach.

svasek:
Sorry,

I don’t understand how the integration could be installed into the `luxtronik2` folder if it is stored in `custom_components/luxtronik`, even in the release archive.

It would probably only be installed into `custom_components/luxtronik2` if another integration was already installed in `custom_components/luxtronik`.

So, there will likely be many users who have it installed in `custom_components/luxtronik`.

Anyway, the automatic deletion of the `luxtronik` folder is applied only if `luxtronik/manifest.json` contains `"domain": "luxtronik2"`.

svasek:
Well, I understand now!

I dug into the HACS source code to understand how this works in practice.

**HACS always installs based on the `domain` from `manifest.json`, not the folder name in the repository.**


The local installation path is determined by `localpath` in [`integration.py`](https://github.com/hacs/integration/blob/main/custom_components/hacs/repositories/integration.py):

```python
@property
def localpath(self):
    return f"{self.hacs.core.config_path}/custom_components/{self.data.domain}"
```

Where `self.data.domain` comes from reading `manifest.json`:
```python
self.data.domain = manifest["domain"]
```

So even though the repo has the code in `custom_components/luxtronik/`, HACS reads `"domain": "luxtronik2"` from the manifest and installs everything into `custom_components/luxtronik2/` on the user's machine.

**This works for both download methods:**

1. **Standard GitHub archive** (what we use) — HACS strips the remote path (`custom_components/luxtronik/`) from each file in the archive and extracts into the local path (`custom_components/luxtronik2/`). The remapping happens in `download_repository_zip()` in [`base.py`](https://github.com/hacs/integration/blob/main/custom_components/hacs/repositories/base.py).

2. **Custom release ZIP** (`zip_release: true` in `hacs.json`) — the ZIP is extracted directly into `custom_components/{domain}/`, so the ZIP contents must be at root level (no subdirectory).

**What this means for renaming the folder:**

Renaming `custom_components/luxtronik/` → `custom_components/luxtronik2/` in the repo aligns the repository structure with what HACS already does on the user's machine. For HACS users, nothing changes at all — they already have `custom_components/luxtronik2/`. The rename just makes the repo less confusing and eliminates the implicit path remapping.

Manual installation users would need to rename their folder (or delete and re-copy), but that's a small audience.


So I can probably remove the automatic deletion from the PR, even though it’s safe.

svasek:
I’ve just removed the automatic cleanup of the old folder from the PR.

AJediIAm:
> I don’t understand how the integration could be installed into the `luxtronik2` folder if it is stored in `custom_components/luxtronik`, even in the release archive.
> 
> It would probably only be installed into `custom_components/luxtronik2` if another integration was already installed in `custom_components/luxtronik`.

Do I assume correctly that you don't have this integration installed yourself? 
(No judgement. Just want to understand how to approach a review. We're very happy with your contributions.)

svasek:
> Do I assume correctly that you don't have this integration installed yourself? (No judgement. Just want to understand how to approach a review. We're very happy with your contributions.)

I've been using this integration for about two years, which is why I eventually started contributing.

I hadn't checked the installation directory on my system until you mentioned it - sorry for not paying attention. But yes, it is indeed installed under the `luxtronik2` folder, I can confirm that.

AJediIAm:
> > Do I assume correctly that you don't have this integration installed yourself? (No judgement. Just want to understand how to approach a review. We're very happy with your contributions.)
> 
> I've been using this integration for about two years, which is why I eventually started contributing.
> 
> I hadn't checked the installation directory on my system until you mentioned it - sorry for not paying attention. But yes, it is indeed installed under the `luxtronik2` folder, I can confirm that.

No worries. Even the best can overlook a simple detail once in a while 😉
That's why we work together on this. Assumptions lead to interesting discussions. I never realized how counter initiative some of the things.

svasek:
Anyway, my primary motivation was the test framework incompatibility.

I started working on tests, but `pytest-homeassistant-custom-component` expects the folder name to match the domain when using `hass.config_entries.async_setup()`, which makes writing full integration-level tests almost impossible.

That, and a little bit of OCD - I just need things to be done properly. 😄

AJediIAm:
@Svasek : can you have a look at the conflicts?

svasek:
> can you have a look at the conflicts?

Rebased to the latest `main` to resolve conflicts.
I've also fixed the linter issues.

Issue #455, opened 2025-10-06
Hassfest error: [MANIFEST] Domain does not match dir name


rhammen:
successor of #390, only for error: [MANIFEST] Domain does not match dir name
see [workflow action logs](https://github.com/BenPru/luxtronik/actions/runs/18265289035/job/51998730285#step:3:73)

rhammen:
Should be fixed with #598 !

rhammen:
Fixed with #598 !

Issue #165, opened 2023-10-21
Provide this integration in HACS


BenPru:
At the moment this integration is only available if you add it a custom repo in HACS.
It would be nice if HACS know it without custom repo.

rhammen:
Documentation how to add to HACS is here: https://hacs.xyz/docs/publisher/
I created a PR to add the 2 needed workflow actions.
The workflow action still shows some things that must be resolved.

rhammen:
HACS validation passes.
With hassfest, biggest problem seems to be "Error: R] [MANIFEST] Domain does not match dir name"
This is about integration name "luxtronik" versus dirname "luxronik2" ...

rhammen:
There are several problems with the translation keys, e.g. :
Invalid translation key 'heatpump running', need to be [a-z0-9-_]+ and cannot start or end with a hyphen or underscore. for dictionary value @ data['entity']['sensor']['status_line_1']['state']. Got {'heatpump running': 'The heatpump runs', 'heatpump idle': 'The heatpump is idle', 'heatpump coming': 'The heatpump is coming', 'heatpump shutdown': 'The heatpump is shuting down', 'errorcode slot': 'Heatpump error code in slot 0', 'pump forerun': 'Pump forerun', 'defrost': 'Defrost', 'writing on LIN connection': 'Writing on LIN connection', 'compressor heating up': 'The Compressor is heating up', 'compressor heater': 'Compressor heater'}

So spaces are not allowed. Anything against changing all spaces in the keys (e.g. in class LuxOperationMode(StrEnum) ) to underscores?

rhammen:
> But wait... Bounis Luxtronik is available in HACS. How does he fullfil the requirements?

He does not use translations, so no issue with the translation keys.

BenPru:
> There are several problems with the translation keys

Where can I see the log for the problems?

rhammen:
> > There are several problems with the translation keys
> 
> Where can I see the log for the problems?

Just go to the 'Actions' tab in the repository. (It's to the right of the 'Pull requests' tab).

rhammen:
Reading the [translations documentation](https://developers.home-assistant.io/docs/internationalization/core#entities), isn't it as simple as adding a translation_key to the entity definitions?
I see several enityties with "a translation_key_name" property instead of a "translation_key" property.
And several entities lack a translation_key property.
I'll s have a look if I can fix the translation errors in this way.

BenPru:
This message is wrong:

[...]

PR #608, opened 2026-05-18
fix: 🐛 revert folder rename to fix HACS installation

### LINKED ISSUE #605: Hacs error: custom_components/luxtronik/manifest.json
Left over bug from #598 

Image

svasek:
### 🐛 Problem

HACS caches `content.path.remote` (e.g. `"custom_components/luxtronik"`) after a repository is first added. Once cached, HACS never re-discovers the folder — it uses the stored path directly to locate `manifest.json`.

After PR #598 renamed `custom_components/luxtronik/` → `custom_components/luxtronik2/`, HACS kept looking at the old path → **manifest not found → installation/update fails**.

Relevant HACS source: [`repositories/integration.py`](https://github.com/hacs/integration/blob/2.0.5/custom_components/hacs/repositories/integration.py) — the `get_integration_manifest` method only re-scans when `content.path.remote == "custom_components"` (initial default).

### ❌ Alternatives considered

- **`zip_release` alone** — HACS still validates manifest via git tree at cached path before downloading
- **Symlink** — GitHub API doesn't expose symlinks as directories
- **Stub folder + `zip_release`** — viable (stub manifest at old path satisfies HACS discovery, zip flat files get installed to correct domain folder), but adds maintenance complexity
- **Remove & re-add in HACS** — too much user friction

### ✅ Solution

Revert the folder name back to `custom_components/luxtronik/`. This is safe because HA tolerates folder name ≠ domain (`luxtronik2`).

### 📝 Changes

- 🔄 Rename `custom_components/luxtronik2/` → `custom_components/luxtronik/`
- 🧪 Update all test import paths (`custom_components.luxtronik2` → `custom_components.luxtronik`)
- 👷 Update CI workflow paths (linter, typecheck, test)
- 🔧 Update `pyproject.toml` coverage path
- 🎨 Fix pre-existing ruff import-sort issue in `model.py`

### 🧪 Verification

- ✅ 638 tests passing
- ✅ `ruff check` — 0 errors
- ✅ `ruff format --check` — all formatted
- ✅ `basedpyright` — 0 errors, 0 warnings

Resolves #605


rhammen:
@Svasek @AJediIAm 
Bummer....
Problem with reverting is that we get #455 back.
And #455 prevents us from implementing #165 , which is a long-term goal....
So this sounds like we (forever?) are abandoning that goal, which I do not like.

Is "Remove & re-add in HACS" really not an option?
Any other ideas are also welcome...

svasek:
You're right, I didn't consider #455 / #165 — reverting the folder would permanently block that path. Let me reconsider.

I traced the full call chain in HACS source ([v2.0.5](https://github.com/hacs/integration/tree/2.0.5)) and found that the fix is much simpler than expected.

## Root cause: `content.path.remote` is ephemeral (in-memory only)

HACS persists repository data in `.storage/hacs.repositories`, but `content.path.remote` is **not** among the exported fields:

```python
EXPORTED_DOWNLOADED_REPOSITORY_DATA = EXPORTED_REPOSITORY_DATA + (
    ("archived", False), ("config_flow", False), ("installed", False),
    ("installed_commit", None), ("last_version", None), ...
    # no content_path_remote!
)
```

On every HA startup, [`HacsIntegrationRepository.__init__()`](https://github.com/hacs/integration/blob/2.0.5/custom_components/hacs/repositories/integration.py) sets the default:

```python
self.content.path.remote = "custom_components"
```

And [`async_restore_repository()`](https://github.com/hacs/integration/blob/2.0.5/custom_components/hacs/utils/data.py) restores `installed`, `domain`, `last_version`, etc. from storage — but **never touches `content.path.remote`**. It stays at the default `"custom_components"`.

Then on the first `update_repository()` call, re-discovery kicks in:

```python
if self.content.path.remote == "custom_components":
    name = get_first_directory_in_directory(self.tree, "custom_components")
    self.content.path.remote = f"custom_components/{name}"
```

After the rename, this finds `luxtronik2` → `content.path.remote = "custom_components/luxtronik2"` ✓

## Why the error occurred

The cached path only lives in memory for the lifetime of the HACS object. The error in #605 happened because the user **had not restarted HA** since the folder rename was published:

1. Before rename: HACS discovered `custom_components/luxtronik` → cached in memory
2. PR #598 renamed folder to `custom_components/luxtronik2`
3. Without HA restart, cached path still points to old `custom_components/luxtronik`
4. User tries to update → `manifest_path = "custom_components/luxtronik/manifest.json"` → not found → error

## The fix: restart Home Assistant

After HA restart:

1. `content.path.remote` resets to `"custom_components"` (default from `__init__`)
2. First `update_repository()` re-discovers → finds `custom_components/luxtronik2` ✓
3. Everything works normally

When the user opens the repo in the HACS UI, this triggers `update_repository` automatically:

```python
if not repository.updated_info:
    await repository.update_repository(ignore_issues=True, force=True)
    repository.updated_info = True
```

No "remove & re-add" needed. No migration steps. Just restart.

## HACS update + changelog visibility — verified ✅

**Version detection is independent of the manifest.** `common_update_data()` fetches releases via GitHub API and sets `data.last_version` **before** the manifest is read. When `async_get_integration_manifest()` raises `HacsException` because the cached path is wrong, the exception is caught by [`QueueManager.execute()`](https://github.com/hacs/integration/blob/2.0.5/custom_components/hacs/utils/queue_manager.py) which uses `asyncio.gather(*tasks, return_exceptions=True)`. After the queue finishes, `data.async_write()` persists the updated version to disk.

**Result:** Users **will see the update notification and release notes** in HACS, even with the stale cached path. The error only blocks the actual install/download — and goes away after a restart.

## Next step: verify the restart fix

The analysis above is based on reading the HACS v2.0.5 source code. Before we close PR #608 and commit to this approach, it would be good to have someone **actually test it** — restart HA and confirm that HACS re-discovers `custom_components/luxtronik2` and the update works without errors.

@AJediIAm — could you try restarting Home Assistant and then check if the HACS error goes away? If it does, no further action is needed on your side.

## Suggested approach (pending verification)

1. **Keep the folder as `custom_components/luxtronik2/`** (PR #598 stays, #455 stays fixed, path to #165 remains open)
2. **Close PR #608** (revert is not needed)
3. **Add a note to the release notes:**
   > ⚠️ If HACS shows an error about `custom_components/luxtronik/manifest.json`, simply **restart Home Assistant**. The integration folder was renamed and HACS needs a restart to re-discover the new path. This is a one-time issue.
4. Reply to #605 with the explanation and close it (after restart is confirmed to work)

svasek:
> Is "Remove & re-add in HACS" really not an option? Any other ideas are also welcome...

TL;DR: @AJediIAm Could you test this? Try to download/update in HACS (you'll get the manifest error), then restart Home Assistant and try to download/update again. The cached path resets on restart, so it should work on the second attempt.

AJediIAm:
> > Is "Remove & re-add in HACS" really not an option? Any other ideas are also welcome...
> 
> TL;DR: @AJediIAm Could you test this? Try to download/update in HACS (you'll get the manifest error), then restart Home Assistant and try to download/update again. The cached path resets on restart, so it should work on the second attempt.

I'll try it this evening.

AJediIAm:
A restart fixes the issue.
It's a bit of an awkward change, but I think it's acceptable to help us move in the right direction.

svasek:
I've tried to download the new pre-release and it didn't work for me even after the HA restart.

## Update: restart alone doesn't fix this

I used AI (Claude Opus 4.6) to analyze the [HACS v2.0.5 source](https://github.com/hacs/integration/tree/2.0.5) and found that the cached path (`content.path.remote`) is **not persisted** — it resets to `"custom_components"` on every HA restart. So in theory, a restart should trigger re-discovery.

**However**, the re-discovery uses the tree for the version returned by [`version_to_download()`](https://github.com/hacs/integration/blob/2.0.5/custom_components/hacs/repositories/base.py), which is the **latest non-prerelease** release. If that release is from before the folder rename, HACS re-discovers the **old** path — even after a restart:

```python
# common_update_data() — determines which tree to scan:
for release in releases:
    if release.prerelease:
        self.data.prerelease = release.tag_name
    else:
        self.data.last_version = release.tag_name  # ← first non-prerelease
        break

self.ref = self.version_to_download()  # → data.last_version (old release)
self.tree = await self.get_tree(self.ref)  # → tree with custom_components/luxtronik/
```

```python
# update_repository() in integration.py — path discovery:
if self.content.path.remote == "custom_components":
    name = get_first_directory_in_directory(self.tree, "custom_components")
    self.content.path.remote = f"custom_components/{name}"  # → "custom_components/luxtronik"
```

Once set, the path is **never re-evaluated** (the `==` check prevents it). When the user then downloads a newer prerelease version that has `custom_components/luxtronik2/`, HACS still looks for `manifest.json` at the old path → error.

### The fix

**Create a new regular (non-prerelease) release.** This makes `version_to_download()` return a version whose tree has `custom_components/luxtronik2/`. After that, users need to **restart HA** so the cached path resets and re-discovery picks up the new folder.

For users who can't wait: remove and re-add the repository in HACS (HACS → ⋮ → Custom repositories → remove → re-add). This creates a fresh instance with the default path and triggers re-discovery.

### Release notes suggestion

> ⚠️ If HACS shows an error about `custom_components/luxtronik/manifest.json`, **restart Home Assistant**. The integration folder was renamed and HACS needs a restart to re-discover the new path. This is a one-time issue.
>
> If the error persists, remove and re-add the repository in HACS: HACS → ⋮ → Custom repositories → remove `BenPru/luxtronik` → re-add `https://github.com/BenPru/luxtronik` (Integration).

PR #613, opened 2026-05-19
fix(ci): 🐛 fix release workflow

svasek:
### ✨ Improvements

- **Add same-day version collision handling** — if the workflow is triggered multiple times on the same day, it now auto-increments a suffix (`2026.05.19` → `2026.05.19.1` → `2026.05.19.2`) instead of failing on duplicate tag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,9 +18,15 @@ jobs:
       - name: Generate Version Number
         id: version
         run: |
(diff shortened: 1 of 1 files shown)


AJediIAm:
It's not common to have multiple increments on one day. 
In the rare case a second release is needed, the next day is commonly chosen. 
The 4th number breaks the hard coded date recognition we have in our brain.

PR #615, opened 2026-05-19
Combine hacs and hassfest actions

AJediIAm:
Since both are required for HACS, both actions are combines.


AJediIAm:
We're getting close...

PR #622, opened 2026-05-22
Fix #621 increase range for cooling thresholds

### LINKED ISSUE #621: Cooling outdoor temp threshold has new upper and lower values
The COOLING_OUTDOOR_TEMP_THRESHOLD had a previous lower value of 18 and upper of 30. After a Luxtronik firmware update this has been changed to a lower of 10 and an upper of 35.

Image

Image

I'll create a PR to update this.

swennijboer:
Fix for [#621](https://github.com/BenPru/luxtronik/issues/621) to align the min/max of the cooling threshold with the current state of the firmware.


AJediIAm:
This change only applies to specific firmware versions.

rhammen:
> This change only applies to specific firmware versions.

@AJediIAm @swennijboer 
Do you know from which firmware version the new limits apply, or do you have a suggestion for that?
I can confirm that in 3.92.1 the new limits (10.0 - 35.0) apply.

AJediIAm:
I couldn't find any specific reference. 3.92.1 seems fine to me.

swennijboer:
I only started noticing it in 3.92.2.

rhammen:
@swennijboer 
Can you then update the PR, so that :
- there is a LuxtronikNumberDescription for SensorKey.COOLING_OUTDOOR_TEMP_THRESHOLD, with the new values, including  
min_firmware_version=Version("3.92.1"),
in the description.
- there is an additional LuxtronikNumberDescription for SensorKey.COOLING_OUTDOOR_TEMP_THRESHOLD, with the old values, including  
max_firmware_version=Version("3.92.0"),
in the description.

This is the same approach as we used for the LuxtronikNumberDescription  with key=SensorKey.DHW_TARGET_TEMPERATURE in the same file.

swennijboer:
@rhammen Thanks for the clear instruction and example, I added a minimal version for the new threshold and kept the old thresholds with a max version.

Issue #541, opened 2026-05-02
Cooling Thermostat controlling minimum outdoor temp instead of desired room temp


Huipie1:
The cooling thermostat is controlling the minimum outdoor temperature to release the cooling. A thermostat is normally controlling the desired temperature so this is confusing behaviour.

In climate.py, should the luxtronik_key_target_temperature be changed to LuxParameter.P1148_HEATING_TARGET_TEMP_ROOM_THERMOSTAT or C0228_ROOM_THERMOSTAT_TEMPERATURE_TARGET ? or possibly: LuxParameter.P0132_COOLING_TARGET_TEMPERATURE_MK1

`    LuxtronikClimateDescription(
        key=SensorKey.COOLING,
        hvac_modes=[HVACMode.COOL, HVACMode.OFF],
        hvac_mode_mapping=HVAC_MODE_MAPPING_COOL,
        hvac_action_mapping=HVAC_ACTION_MAPPING_COOL,
        preset_modes=[PRESET_NONE],
        supported_features=ClimateEntityFeature.TURN_OFF
        | ClimateEntityFeature.TURN_ON  # noqa: W503
        | ClimateEntityFeature.TARGET_TEMPERATURE,  # noqa: W503
        luxtronik_key=LuxParameter.P0108_MODE_COOLING,
        luxtronik_key_target_temperature=LuxParameter.P0110_COOLING_OUTDOOR_TEMP_THRESHOLD,
        luxtronik_key_current_action=LuxCalculation.C0080_STATUS,
        luxtronik_action_active=LuxOperationMode.cooling.value,
        # luxtronik_key_target_temperature_high=LuxParameter,
        # luxtronik_key_target_temperature_low=LuxParameter,
        icon_by_state=LUX_STATE_ICON_MAP_COOL,
        temperature_unit=UnitOfTemperature.CELSIUS,
        # visibility=LuxVisibility.V0005_COOLING,
        device_key=DeviceKey.cooling,
    ),
`

rhammen:
I can understand that you find it confusing, but ....
... this is how the luxtronik heatpump control logic works: cooling is enabled when the outdoor temperature exceeds a certain threshold value.
Since we are not building our own control logic, but just re-using the luxtronik control logic, this is why the thermostat controls the (outdoor temperture) threshold value.

[...]

Issue #491, opened 2025-11-24
Feedback request: does changing the Heating Target temperature (using the Heat thermostat tile) work for you?


rhammen:
This is a request to all users of this integration.

With #346 , the implementation was changed such that changing the target room temperature using the thermostat, should change the target luxtronik room temperature. (parameter 1148, new in approx. firmware v3.90)
Goal of this is to be able to change the target room temperature using this integration.
If this works correctly, then when you increase the target room temperature using the thermostat card, the target return flow temperature should also increase.

My question is: does this work for you?

Why do I ask:
For me, it does not work. :-(
I can change the Heating Target Room Temperature using the thermostat tile to e.g. 30 or 35 C, but the return flow temperature does not change at all. When I change the target room temperature on the RBE itself, the target return flow temperture does change correctly.
So for me, changing the parameter 1148 does not result in the target room temperature on the RBE being changed.

Edeheij:
Unfortunately, it doesn't work here either. It would be great if it did. When i set a temp, it goes back to 0C.

AJediIAm:
I did some tests and it looks like the heating target works as intended.

<img width="1008" height="1029" alt="Image" src="https://github.com/user-attachments/assets/d7e15093-c0da-4e47-ab7f-520f4b0e81e4" />

On wednesday, you can see how the heat flow in is opposite from the measured temperature. I assume that this is because of the room impact factor. 
Maybe I should make a slight adjustment to the heatcurve to dail in the temperature a bit better.

On thursday, I lowered the room thermostat target via home assistant. I made a large change to  ensure the effects would be observable. I observed that the physical room thermostat (RBE+) adopted the new temperature as well. 
The heat flow in target adjusted accordingly. Because there was enough heat in the house, the heatpump didn't run that day. Also the wife confirmed the temperature had dropped at the end of the day, so I had to raise the temperature a bit again.

On friday, the heatpump had to run for a bit. Besides the idle time on thurday, I didn't see a notable change in the power consumption. The heat flow in target is enough evidence for me that the temperature change had the desired effect.

kwisse:
I'm using number.luxtronik_230305_047_heating_target_correction for correcting the target temperature.
My thermostat does not set a temperature request, but an offset on the heating curves.
The above parameter is another offset on the heating curves so the physical jog wheel still works.

AJediIAm:
I tried the target correction and noticed a different effect. Instead of affecting the heat flow in, it only affects the heat flow out. 

[...]

PR #630, opened 2026-05-30
Add RoomThermostat detection and differentiates Climate thermostats based on this

rhammen:
- Add Room thermostat type detection. Type is for now logged in logging.
- Differentiates thermostats based on this:

Thermostat == Smart:
- Heating changes P1148 as before
- Cooling now also changes P1148, instead of outside temperature threshold. This addressed #541

Thermostat != Smart:
- Heating: now changes the traget temperature correction factor [-5.0C , +5.0C] , because changing P1148 only seems to work for smart thermostats, see #491
- Cooling still changes outside temperature threshold.


AJediIAm:
Something is not right.

The heating control climate entity has 0C as target while Cooling has 21C as target. My heatpump has been mostly idle this week, so I have done little testing. 

See https://github.com/BenPru/luxtronik/issues/639 for details

Issue #631, opened 2026-05-31
Update icons


AJediIAm:
While checking https://github.com/BenPru/luxtronik/pull/628, I noticed some of the icons didn't match the manual/display. 

I suggest to make the changes shown in bold.

Status:
- heating: "mdi:radiator", 
- **hot_water: "mdi:water-pump",**
- swimming_pool_solar: "mdi:pool",
- **evu: "mdi:solar-power",**
- **defrost: "mdi:snowflake-melt",**
- no_request: "mdi:hvac-off",
- **heating_external_source: "mdi:water-boiler",**
- cooling: "mdi:snowflake" <img width="305" height="42" alt="Image" src="https://github.com/user-attachments/assets/b64845a0-7382-464d-acb6-0b2ccf51a831" />

AJediIAm:
Cooling:

Snowflake is also used by Luxtronik, so that's the main theme here. 
<img width="304" height="332" alt="Image" src="https://github.com/user-attachments/assets/f6fb7b09-2773-4fad-b0c5-743ec35e27ad" />

I seem to be something wrong with the state-based icons. 
The icon doesn't change when it is turned of. 
<img width="327" height="49" alt="Image" src="https://github.com/user-attachments/assets/9ea2893d-42ff-4636-9924-807aa5fb27a0" />
Any suggestions?

svasek:
Hi @AJediIAm,

There is no state icon defined for the `cooling` switch (line `158`), so it falls back to the default icon.
What icon would you like to use for the `on` and `off` states? I guess you want to use the same as for `climate` (line `219`).

AJediIAm:
Correct. Snowflake and snowflake-off. 
I ran out of free tokens trying to setup ha-mcp this weekend, so I'm stuck with updates by hand. Who does that these days...

svasek:
Just replace definition for `cooling` switch on line `158` with this:

```json
      "cooling": { 
        "default": "mdi:snowflake" },
        "state": {
          "on": "mdi:snowflake",
          "off": "mdi:snowflake-off"
        }
```

Also, it's better to always define a default icon.

AJediIAm:
Initial update done in https://github.com/BenPru/luxtronik/pull/632 
There are still a few inconsistencies to be fixed.

AJediIAm:
Closing the issue for now

PR #632, opened 2026-05-31
Update icons

AJediIAm:
Update status icons to better match luxtronik icons.
Part of https://github.com/BenPru/luxtronik/issues/631

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,17 +127,21 @@ Cooling is controled using the "cooling" [climate entity](https://www.home-assis
 
 <img width="174" height="102" alt="image" src="https://github.com/user-attachments/assets/a935b390-6c9d-47d4-8173-c9233ed07e2f" />
 
-The climate entity allows you to control the state. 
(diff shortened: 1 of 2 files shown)


AJediIAm:
It appears the cooling device names it's entities with a different prefix, resulting in it ignoring the specified icon. It's a guess, so maybe it's something else. 
image

svasek:
Hi, @AJediIAm, 

For this PR, I would suggest manually squashing all commits in this branch into one, or using the `Squash and merge` button instead of `Merge` as there are actually 8 commits in the branch.

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges

AJediIAm:
Thanks for the tip

rhammen:
@AJediIAm Not sure if you are already finished, I assume you will merge this yourself when you are done.

AJediIAm:
I need to fix a few bugs.

PR #633, opened 2026-05-31
Added an option to set the minimal cooling temperature value

Jaarden:
I was missing the option to change the minimal cooling temperature value, setting this value allows you to have better dew control and in or decrease your floor cooling temperatures. Added the option to not set it below 17.0c as a failsafe, because this can cause some wet and slippery floors on a hot summer day (speaking of experience, and i can not recommend).


rhammen:
The convention is that in the LuxParameter class, we start the name with P<parameter_number>.
You used:
    P0852_COOLING_MIN_FLOW_OUT_TEMPERATURE = "parameters.ID_Einst_min_VL_Kuehl"
This suggests that ID_Einst_min_VL_Kuehl has parameter number 852.

However, in the Bouni Luxtronik library that we use under water, parameter 852 = "ID_Waermemenge_Seit".
And parameter "ID_Einst_min_VL_Kuehl" has number 993...

So I would have expected:
    P0993_COOLING_MIN_FLOW_OUT_TEMPERATURE = "parameters.ID_Einst_min_VL_Kuehl"

Can you correct this please?

Jaarden:
Sorry, was not aware of this. But i fixed it.

PR #635, opened 2026-06-01
feat: ✨ add repair issues for connection failure and unsupported firmware

svasek:
### ✨ Summary

Implement the [Integration Quality Scale `repair-issues` rule](https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/repair-issues/) (Gold tier) by creating a repair issue when the integration fails to connect to the heat pump during setup.

### 🐛 Problem

When the integration fails to connect, the user only sees a generic "not ready" state in the UI with no actionable guidance. The connection failure is recoverable but requires user intervention (check network, verify host/port, ensure heat pump is powered on).

#### Repair issue (`__init__.py`)

- 🔌 **`connection_failed`** — created when `async_setup_entry` fails to connect (severity: error)
  - Includes host, port, and error message in the issue description
  - Automatically cleared on the next successful setup
  - Issue ID scoped per config entry (`connection_failed_{entry_id}`) for multi-instance support
  - `is_fixable=False` (user resolves externally), `is_persistent=False` (re-evaluated each setup)

#### Translations

- 🌍 Add `issues.connection_failed` (title + description) to all 5 languages (en, de, cs, nl, pl)
- Description includes placeholders for `{host}`, `{port}`, and `{error}` with guidance to use the Reconfigure option

### ✅ Tests

- ➕ `test_connection_failure_creates_repair_issue` — verifies `ir.async_create_issue` called with correct key, severity, and placeholders on connection error
- ➕ `test_successful_setup_clears_connection_issue` — verifies `ir.async_delete_issue` called for `connection_failed` on success

### 📦 Files changed

| File                                                | Change                                   |
| --------------------------------------------------- | ---------------------------------------- |
| `custom_components/luxtronik2/__init__.py`          | add repair issue creation/deletion logic |
| `custom_components/luxtronik2/translations/en.json` | add `issues` section                     |
| `custom_components/luxtronik2/translations/de.json` | add `issues` section                     |
| `custom_components/luxtronik2/translations/cs.json` | add `issues` section                     |
| `custom_components/luxtronik2/translations/nl.json` | add `issues` section                     |
| `custom_components/luxtronik2/translations/pl.json` | add `issues` section                     |
| `tests/test_init.py`                                | 2 new tests for connection failure issue |

### 🧪 Verification

- ✅ 651 tests passing
- ✅ basedpyright: 0 errors
- ✅ ruff check + format: clean
- ✅ `__init__.py`: 100% coverage

### DIFF
--- a/custom_components/luxtronik2/__init__.py
+++ b/custom_components/luxtronik2/__init__.py
@@ -9,7 +9,7 @@ from homeassistant.config_entries import ConfigEntry, ConfigEntryState
 from homeassistant.const import CONF_HOST, CONF_PORT, CONF_TIMEOUT, Platform as P
 from homeassistant.core import HomeAssistant
 from homeassistant.exceptions import ConfigEntryNotReady, ServiceValidationError
-from homeassistant.helpers import device_registry as dr
+from homeassistant.helpers import device_registry as dr, issue_registry as ir
 from homeassistant.helpers.entity_registry import (
     async_get,
 )
@@ -47,7 +47,24 @@ async def async_setup_entry(hass: HomeAssistant, entry: LuxtronikConfigEntry) ->
         await coordinator.async_config_entry_first_refresh()
     except Exception as err:
         LOGGER.error("Luxtronik connection failed: %s", err)
+        ir.async_create_issue(
+            hass,
+            DOMAIN,
+            f"connection_failed_{entry.entry_id}",
+            is_fixable=False,
+            is_persistent=False,
+            severity=ir.IssueSeverity.ERROR,
+            translation_key="connection_failed",
+            translation_placeholders={
+                "host": str(config.get(CONF_HOST, "unknown")),
+                "port": str(config.get(CONF_PORT, "")),
+                "error": str(err),
+            },
+        )
         raise ConfigEntryNotReady from err
+
+    # Clear any previous connection failure issue
+    ir.async_delete_issue(hass, DOMAIN, f"connection_failed_{entry.entry_id}")
 
     entry.async_on_unload(entry.add_update_listener(update_listener))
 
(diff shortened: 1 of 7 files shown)


AJediIAm (approved):
Looks good.

rhammen:
@Svasek @AJediIAm 
I'm concerned about the unsupported_firmware_version check.
We have a lot of different heatpumps manufacturers/models.
If the check fails, then why would an updated firmware version fix it? The updated firmware version may well have the same format, and thus result in the same error... So is it really user-fixable?
There may be cases where such an error requires changes to the firmware version parsing first... Without the integration starting  up, I'm not sure users then even have the possibility to share a diagnostics file for use to diagnose?

If users cannot configure the integration when the firmware version is not recognized for their heatpump manufacturer, I'm afraid we get support requests of users who upgraded the integration and suddenly cannot use it anymore...

What is your view on this?

svasek:
Hi @rhammen,

You raise a very valid point — I've removed the `unsupported_firmware_version` repair issue entirely.

You're right that an unparseable firmware version format is more likely a parsing issue in the integration code than something the user can fix by updating firmware. A newer firmware version from the same manufacturer would likely use the same format, so the repair issue would just alarm users without giving them any actionable resolution. And since the integration still starts up fine with `Version("0")` (only some entity visibility is affected via `min/max_firmware_version` filtering), blocking users with a repair warning would generate unnecessary support requests.

The existing `LOGGER.warning` in the coordinator is sufficient for diagnostics — and since setup completes normally, users can still share diagnostics files for us to investigate the version format.

The PR now only contains the `connection_failed` repair issue, which is genuinely user-fixable (check network, power, host/port settings).

PR #640, opened 2026-06-07
Enhance Room Thermostat Type detection

### LINKED ISSUE #639: Cooling climate control bug in 2026.6.6
Something is not right

This is the previous release:
image

This is the new release:
image

Note that the heating control climate entity has 0C as target while Cooling has 21C as target. My heatpump has been mostly idle this week, so I have done little testing.

rhammen:
Trying to fix #639, by detecting RBE+ based on RBE firmware version

### DIFF
--- a/custom_components/luxtronik2/climate.py
+++ b/custom_components/luxtronik2/climate.py
@@ -206,7 +206,10 @@ async def async_setup_entry(
     rt = getattr(coordinator, "room_thermostat_type", None)
     is_smart_thermostat = False
     if isinstance(rt, LuxRoomThermostatType):
-        is_smart_thermostat = rt == LuxRoomThermostatType.smart
+        is_smart_thermostat = rt in (
+            LuxRoomThermostatType.smart,
+            LuxRoomThermostatType.rbe_plus,
+        )
 
     LOGGER.info(
         "Detected room thermostat type: %s (smart=%s)",
--- a/custom_components/luxtronik2/const.py
+++ b/custom_components/luxtronik2/const.py
@@ -229,6 +229,7 @@ class LuxRoomThermostatType(Enum):
     rfv_dk = 3
     rbe = 4
     smart = 5
+    rbe_plus = 90
 
 
 class LuxSwitchoffReason(Enum):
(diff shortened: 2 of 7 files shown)


AJediIAm:
It now correctly updates and syncs with HEATING_TARGET_TEMP_ROOM_THERMOSTAT.
Can you merge it to main when the tests are fixed?
If you want, you can run the "create hacs pre-release" action. Otherwise I'll run it after you merged it. 

As a follow-up change, would it be possible to combine the heating climate entity and cooling climate entity into 1 central heatpump climate entity? I believe home assistant supports this. Any thoughts on this?

rhammen:
@AJediIAm 

> If you want, you can run the "create hacs pre-release" action. Otherwise I'll run it after you merged it.

Done!

> As a follow-up change, would it be possible to combine the heating climate entity and cooling climate entity into 1 central heatpump climate entity? I believe home assistant supports this. Any thoughts on this?

I personally have no interest in doing such change.:
I don't have a RBE+, and I thus need separate target temperature control for heating and cooling. Thus 2 climate entities.
I also wouldn't be able to test such a change to 1 climate entity myself, so I will not implement such changes.

I can understand that users with an RBE+ would prefer to have 1 climate entity.
So please feel free to implement this yourself, but make sure you don't break the integration for users without RBE+ :-)

AJediIAm:
I'm also wondering if it is worth adding complexity to the integration. 
It's good to only show the entities which are relevant for the users firmware and setup. 
It's not good to have the code littered with conditions. 
Is the code good enough for reuse or do you have improvements in mind to make this more widely available?

rhammen:
> I'm also wondering if it is worth adding complexity to the integration. It's good to only show the entities which are relevant for the users firmware and setup. It's not good to have the code littered with conditions. Is the code good enough for reuse or do you have improvements in mind to make this more widely available?

I'm not really sure what your are asking exactly. So I'll try to answer, but maybe you meant something different.

The Room_thermostat_type is present is a property of the coordinator, and in this way easily accessible to the different entity platforms. In that sense it is re-usable.
Currently only the climate platform discriminates based on this.
The only improvement I currently can think of is dealing with the case of No Room thermostat present. Then the climate entities should not have the possibility to set the target temperature...

AJediIAm:
> The Room_thermostat_type is present is a property of the coordinator, and in this way easily accessible to the different entity platforms. In that sense it is re-usable. Currently only the climate platform discriminates based on this. The only improvement I currently can think of is dealing with the case of No Room thermostat present. Then the climate entities should not have the possibility to set the target temperature...

Do you know if the target temperature correction work when there is no room thermostat? 
Maybe we should create a table with the (desired) behavior per setup.

PR #642, opened 2026-06-08
Add selector for parameter 119 (ID_Ba_Sw_akt)

rhammen:
Code tries to discrimnate between Pool mode and PVmode. 
In case Parameter 119 = Automatic during HA startup, it cannot discriminate, and it will allow ALL options (both PV mode and Pool mode).

Additionally, refactored code: 
moved select entity descriptions to separate file select_entities_predefined.py, similar to the other platforms.


rhammen:
@svasek  Would you be willing to help bringing the test-coverage again to 100%? That would be great :-)

PR #647, opened 2026-06-09
Add missing '2nd heatsource' option to DHW select entity

rhammen:


iansenkowalski:
Great and what’s about extra DHW or spore charge?

br

PR #649, opened 2026-06-10
feat(config_flow): broadcast discovery on every enabled IPv4 interface

### LINKED ISSUE #646: Discovery broadcasts only on the default route — fails on multi-homed HA hosts
## Summary

`lux_helper.discover()` sends its magic UDP packet to `255.255.255.255`,
which the kernel routes out the default interface only. On Home Assistant
hosts with multiple network interfaces (extra NICs, VLANs, Docker
bridges), the broadcast never reaches devices on the other subnets, so
"Add Integration → Luxtronik" returns zero hits and the user has to drop
to manual entry.

I'm seeing this on my own multi-VLAN setup: heatpump is reachable from
HA via static route, but the broadcast only goes out the default
interface and misses it.

## Proposed fix

Have `_discover_devices()` enumerate every enabled IPv4 adapter via HA
core's `network.async_get_ipv4_broadcast_addresses` helper and pass the
per-interface broadcast addresses through to `discover()`. The discover
function takes a new optional `broadcast_addresses: list[str] | None`
parameter; default `None` preserves the existing single-broadcast
behavior so external callers (the python-luxtronik library, the CLI
helper) stay unchanged.

Same pattern as HA core's flux_led, wiz, and elkm1 integrations.

## Implementation

I have this ready on a branch — three files, ~95 lines mostly comments
and tests:
https://github.com/mattias-arrelid/luxtronik/compare/main...feat/multi-interface-discovery

- `custom_components/luxtronik2/lux_helper.py` — `discover()` gains the
  optional parameter; ~6 lines of logic change.
- `custom_components/luxtronik2/config_flow.py` — `_discover_devices()`
  fetches broadcasts via the HA helper; ~6 lines.
- `tests/test_lux_helper.py` — 3 new test cases (default-fallback,
  multi-address path, empty-list-fallback). All 57 existing tests still
  pass.

Verified live: discovery finds the heatpump on a VLAN that the kernel's
default route doesn't reach, on both my macOS dev HA and my Raspberry Pi
HA. Backward-compatible — single-NIC users see no behavioral change.

Happy to open a PR if there's interest. Following the project's
gitmoji + Conventional Commit format per the contributor instructions.

mattias-arrelid:
Fixes #646.

Discovery currently broadcasts to `255.255.255.255`, which the kernel routes out the default interface only. On multi-homed Home Assistant hosts (extra NICs, VLANs, Docker bridges) heatpumps on other subnets aren't found.

This enumerates every enabled IPv4 adapter via HA's `network.async_get_ipv4_broadcast_addresses` and sends the magic packet to each. Same pattern as `flux_led`, `wiz`, `elkm1`.

`discover()`'s new `broadcast_addresses` param defaults to `None`, preserving the legacy single-broadcast for external callers.

### DIFF
--- a/custom_components/luxtronik2/config_flow.py
+++ b/custom_components/luxtronik2/config_flow.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 from typing import Any
 
 from homeassistant import config_entries
+from homeassistant.components import network
 from homeassistant.config_entries import ConfigFlowResult
 from homeassistant.const import CONF_HOST, CONF_PORT, CONF_TIMEOUT
 from homeassistant.core import callback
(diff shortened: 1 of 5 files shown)


rhammen:
@mattias-arrelid Thanks!

Issue #519, opened 2026-02-13
HP status idle when doing thermal disinfection


ChrisMisker:
![image](https://github.com/user-attachments/assets/e260a2a5-8f57-4082-ba90-86765471947c)

See the above image. My hp status is idle instead of domestic water when doing the thermal disinfection run. Yellow/orange is domestic water. 

This is annoying, because I use the hp status to set a utility meter to measure my COP when heating and DHW.

rhammen:
Still not working correctly, needs more analysis to (try to) fix

rhammen:
Checked today, and Status did correctly reflect the TDI as Domestic Water:
First heatpump heats to target temperature, then short stop, and then the TDI brings the temperature to the TDI target temperature.

<img width="1625" height="857" alt="Image" src="https://github.com/user-attachments/assets/59bb6d6c-c3ab-43a8-8197-8bad44febe9b" />

@ChrisMisker Can you also check and confirm pre-release [2026.06.17](https://github.com/BenPru/luxtronik/releases/tag/2026.06.17) works Ok regaridng the HP status during TDI?

ChrisMisker:
Dang.. I did a TDI today, but wasn't on 2026.06.17! I'll update and run one again tomorrow.

ChrisMisker:
@rhammen It seems like it worked... I did have 1 minute of "idle", while it was doing TDI.. see the history graph. But I guess this is as good as it gets? The power is measured with a power meter in the "meterkast" (?) :)

<img width="1512" height="949" alt="Image" src="https://github.com/user-attachments/assets/c760a5f9-3b52-4c21-93af-55892947ff67" />
and zoomed in:

<img width="1512" height="949" alt="Image" src="https://github.com/user-attachments/assets/d87f3f35-09ad-433f-8758-25bf23f91b50" />

rhammen:
Thanks for confirming it works!

Yes, that is the same behaviour as I observed on my Homewizard energy meter.
Since in that period you also see Power = 0 W, I think keeping those 2 or 3 minutes as Idle is reasonable.
(I can think of a way to also mark it as Domestic Water using that period, but I don't think we should do that).
So I propose to keep the current implementation.

ChrisMisker:
[...]

PR #650, opened 2026-06-14
Refactor Heatpump Status workarounds & attempt to fix TDI detection

rhammen:
- Heatpump Status workarounds were present in both sensor.py and common,py.  --> Now all are in common.py
- Attempt to fix TDI detection (issue #519)

### DIFF
--- a/custom_components/luxtronik2/common.py
+++ b/custom_components/luxtronik2/common.py
@@ -96,31 +96,26 @@ def get_sensor_data(
         LOGGER.warning("Get_sensor %s (%s) returns None", sensor_id, luxtronik_key)
         return None
     value = sensor.value  # pyright: ignore[reportAttributeAccessIssue]
-    return value if raw_value else correct_key_value(value, coordinator, luxtronik_key)
+    return (
+        value
+        if raw_value
+        else normalize_sensor_value(value, coordinator, luxtronik_key)
+    )
 
 
-def correct_key_value(
+def normalize_sensor_value(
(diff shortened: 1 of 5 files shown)


AJediIAm:
Some heat pumps don't report good status lines. 
My heat pump always displays the same values and I don't think they lend themselves well for internal logic. Would it be possible to use alternatieve properties?

rhammen:
>     * Attempt to fix TDI detection (issue [HP status idle when doing thermal disinfection #519](https://github.com/BenPru/luxtronik/issues/519))

> Some heat pumps don't report good status lines. My heat pump always displays the same values and I don't think they lend themselves well for internal logic. Would it be possible to use alternatieve properties?

My heatpump also display same statusLine1 and statusLine3 values all the time, but the Status work mostly OK.
This attempted fix is for the Status sensor, basically enhancing the exisitng Thermal Desinfection detection fix for the statusLine3 = (always) heating case.
Is that what you mean with "Would it be possible to use alternatieve properties?"

AJediIAm:
I didn't realize it would potentially fix the status lines when looking at the code. If the status lines are fixed, it's all fine. 
Let's try it.

rhammen:
> I didn't realize it would potentially fix the status lines when looking at the code. If the status lines are fixed, it's all fine. Let's try it.

StatusLine1 and StatusLine3 are not fixed (and I have no intention in try to do so now).
Status should be OK.

PR #653, opened 2026-06-17
Add several number entities, fix translations, add translation coverage github action

rhammen:
New entities:
- _extra_dhw_duration
- _extra_dhw_target_temperature
- _dhw_manual_frequency

New:
- Add github action to check for any missing translation keys

Updated:
- fixed several missing translations


AJediIAm:
We should update the readme when adding entities.
Looking forward to testing the change, so I can update the documentation.

AJediIAm:
I can't trigger the extra DHW remotely. 
The option is on the controller itself, but none of the modes seem to activate it.

PR #657, opened 2026-06-28
Create Efficiency Pump Voltage or Percentage based entities

### LINKED ISSUE #595: Efficiency pump nominal scaling seems incorrect
I am using a Luxtronik / AIT Hybrox 8 heat pump with the Home Assistant Luxtronik integration.

The entity:

`number.luxtronik_0_00_efficiency_pump_nominal`

is exposed as a writable number entity with unit `V` and allowed values from `1` to `10`.

However, the scaling appears to be incorrect on my controller:

* setting `1` results in `100%`
* setting `30` results in `3000%`

Values above `1` therefore become unrealistic and are eventually overwritten by the Luxtronik controller itself.

This suggests that the raw value may internally be interpreted as percentage * 100.

At the same time, values below `1` cannot be entered, so I cannot test values like `0.3` or `0.5`.

It therefore looks like one of the following may be wrong:

* scaling
* min/max values
* step size
* or the entity type itself

Additionally, the switch:

`switch.wp_effizienzpumpe`

does not appear to have any effect.

I would expect this switch to enable/disable automatic control of the efficiency pump, but toggling it does not seem to change anything in the Luxtronik controller UI or in the actual pump behavior.

Has anyone successfully used these entities or knows how these values are supposed to work?

rhammen:
Create Efficiency Pump Voltage or Percentage based entities, should fix https://github.com/BenPru/luxtronik/issues/595
- Added option in code to provide formula to determine if entity should be created or not
- Used that feature to create either Efficiency Pump voltage or Efficiency Pump percentage entities
- This should fix  https://github.com/BenPru/luxtronik/issues/595


AJediIAm:
It's a pretty big change. I haven't read all the code. Let's test it in production ;)

rhammen:
> It's a pretty big change. I haven't read all the code. Let's test it in production ;)

Let's first have a couple of people confirm it works OK with the 2026-06.28 pre-release ;)

PR #672, opened 2026-07-04
fix(config_flow): stop DHCP discovery masking already_configured as unknown error

rhammen:
## 🐛 What

- **DHCP AbortFlow masking:** `async_step_dhcp`'s broad `except Exception` was catching the `AbortFlow` legitimately raised by `_abort_if_unique_id_configured` when a previously-configured heat pump gets a new DHCP lease/IP. Instead of the intended graceful "already_configured" abort (which had already updated the entry's host/port), the flow logged a scary "Unhandled DHCP discovery error" traceback and reported `reason="unknown"`. Fixed by re-raising `AbortFlow` before the catch-all, matching the pattern already used in `async_step_reconfigure`.
- **Missing options translation:** the options flow's `update_interval` selector (added for configurable poll interval) had no translation label in any of the 5 locale files, so it rendered as the raw key `update_interval`. Added labels/descriptions to en/de/nl/pl/cs.

## 🧹 Cleanup

- Removed dead class attributes `_discovery_host`, `_discovery_port`, `_title`, `_sensor_prefix` (never read anywhere).
- Removed an unnecessary self-import in `async_get_options_flow`.
- Removed leftover dead `config.step.options` translation blocks (`control_mode_home_assistant`, `ha_sensor_prefix`) in all 5 locales — cruft from before the options flow moved to the top-level `options` translation key; never read by any code path.

## Test plan

- [x] `pytest` — 747 passed
- [x] `ruff check` — clean
- [x] `basedpyright` — 0 errors
- [x] Reproduced the DHCP bug with a scripted repro (mocking `_abort_if_unique_id_configured` to raise `AbortFlow` as real HA does) — confirmed `AbortFlow` now propagates instead of being swallowed.
- [x] Verified all 5 translation JSON files still parse.

### DIFF
--- a/custom_components/luxtronik2/config_flow.py
+++ b/custom_components/luxtronik2/config_flow.py
@@ -46,11 +46,6 @@ class LuxtronikFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
     """Handle a Luxtronik heatpump controller config flow."""
 
     VERSION = CONFIG_ENTRY_VERSION
-    _discovery_host = None
-    _discovery_port = None
-
-    _sensor_prefix = DOMAIN
-    _title = "Luxtronik"
     _all_devices: list[dict[str, Any]] = []
     _available_devices: list[dict[str, Any]] = []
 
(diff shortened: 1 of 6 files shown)


AJediIAm:
I tried to test the auto discovery, but it's not getting discovered anymore in my test instance. There is a good chance it's on my side. @rhammen any issues on your side?

Issue #188, opened 2023-11-02
Create a CoP Sensor


BenPru:
With this native historic values (#187 ) I can create a CoP Sensor with also statistic attributes.

AJediIAm:
A COP would be a better as a blueprint instead of an enhancement to the integration. I'll try to gather the documentation for this and publish an example.

AJediIAm:
This issue matches my goal to improve the documentation. 

I'll start with the code for the automation and then look into making it a real blueprint. The code can be stored as code with a section in the documentation on how to use it.

siedi:
With the latest firmware update (I think it was 3.89.x or so), my alpha innotec got the sensor.luxtronik2_current_power_consumption exposed. I have no external power measurement with KNX or shelly. Just created a helper sensor (template based) with the following code:

```
{% set q = states('sensor.luxtronik2_current_heat_output') | int %}
{% set w = states('sensor.luxtronik2_current_power_consumption') | int %}
{% if q >= 0 and w > 0 %}
{{ (q / w) | round(2) }}
{% else %}
0
{% endif %}
```

At least I got some reasonable values for the COP. Works for me. Could be easily created into the integration directly of course.

AJediIAm:
Can the COP sensors be moved from "sensors" to "diagnostics"?

The DHW COP is only active for a short amount of time. To provide a more meaningful metric, the average COP over the last 24 hours could be taken. I have a template sensor for the 24 hour average COP and it's a useful metric.

rhammen:
Thanks @AJediIAm — my take on both points:

**On the 24h average: I'd like to try the `statistics` integration route first.**

Rather than baking a fixed window into the integration, the HA [`statistics`](https://www.home-assistant.io/integrations/statistics/) integration lets each user derive whatever averaging window they want — daily, weekly, monthly, yearly — from the existing COP sensors, without us picking one for them.

This works well precisely *because* the instantaneous COP sensors go `unavailable` when the pump isn't serving that circuit: `statistics` skips those samples, so the result is the average COP **while actually running**, which is the meaningful number and directly addresses your "DHW COP is only active for a short amount of time" point.

Both COP sensors also already carry `state_class: measurement`, so Home Assistant records long-term statistics for them automatically — a Statistics card with mean per hour/day/month needs no helper at all.

[...]

PR #693, opened 2026-07-18
feat(sensor): ✨ add instantaneous Heating/DHW COP sensors

rhammen:
## ✨ Summary

- Adds two new sensors, **Heating COP** and **DHW COP**, computed live as `current_heat_output / current_power_consumption` — no external power meter (e.g. Shelly plug) required, since both readings are already exposed by this integration.
- Each sensor only reports a value while the heat pump is actually serving that circuit (gated on the live operating status); it goes `unavailable` otherwise rather than showing a stale or misleading ratio from a different mode.
- Enabled by default, regular (non-diagnostic) sensors, `state_class: measurement`, no `device_class`/unit (dimensionless ratio) — won't appear on the Energy dashboard.
- Translated in all 5 supported locales (en/de/nl/pl/cs).

## 🧊 Cooling COP intentionally excluded

A Cooling COP sensor was built and reviewed, but dropped before merge: `current_heat_output` is not the right numerator during a cooling cycle (it's heat output, not cooling output), so the value would likely be either permanently unavailable (negative-numerator guard) or a misleading ~0. Revisiting once it's clear what the heat pump actually reports during cooling on real hardware.

## 🧪 Testing

- 5 new unit tests covering the ratio computation, status gating, and edge cases (zero/negative/missing values, no coordinator data).
- Full suite: 824 passed, 100% coverage on `custom_components.luxtronik2`, `ruff`/`basedpyright`/`codespell` all clean.

## 📝 Background

Requested in #188 and discussed in #325 — several users had been hand-building COP template sensors from `current_heat_output`/`current_power_consumption`; this adds it natively.


AJediIAm:
My power sensor doesn't work. Maybe it's not physically present. Could an option be added for an external power sensor similar to the room thermostat sensor?

rhammen:
@AJediIAm You do have C0257_CURRENT_HEAT_OUTPUT, but you don't have C0268_CURRENT_POWER_CONSUMPTION ?

AJediIAm:
Current power consumption doesn't make any sense. 

Screenshot_20260719-190814

rhammen:
@AJediIAm Almost looks like between 0:00 and 5:30 there is some heatpump action running with a "calculated" power of 6000 W, which should be (close to) 0 W? Anything in your diagnostics file that may explain this?

rhammen:
@AJediIAm Try the https://github.com/BenPru/luxtronik/tree/feat/external-power-sensor-for-cop branch

PR #743, opened 2026-08-11
Fix | Unable to save options when sensor field is empty

mikeborgeld:
**Home Assistant Core version:** 2026.8.1
**Heat pump firmware:** 3.92.3
**Model:** Alpha Innotec WZSV 62K3M

### Bug description
Submitting the options form (Settings → Devices & Services → Luxtronik → Configure) with the
"indoor temperature sensor" and/or "power consumption sensor" fields left empty raises:
`Entity None is neither a valid entity ID nor a valid UUID`. Happens on a fresh config entry,
or any time these optional fields aren't set. Since the whole form fails validation, this also
blocks saving the update interval.

image

### Root cause
In `schema_helper.py`, `build_options_schema()`, both fields use `vol.Optional(key, default=current_value, ...)`.
When `current_value` is `None`, Voluptuous forces the key into the validated output as `None`,
which `EntitySelector` rejects. The update-interval field avoids this because it already falls
back with `default=current_interval or DEFAULT_UPDATE_INTERVAL_OPTION`.

### Fix
Drop `default=` on both entity-selector fields; `description={"suggested_value": ...}` already
handles the UI pre-fill independently, so nothing is lost.

### Testing
- Empty fields now save without error; `options` correctly omits the keys entirely.
- Clearing a previously-set sensor and saving stores an explicit `null`, which `climate.py`
  and `sensor.py` both already handle correctly.
- Existing valid selections are unaffected.

### DIFF
--- a/custom_components/luxtronik2/schema_helper.py
+++ b/custom_components/luxtronik2/schema_helper.py
@@ -46,7 +46,6 @@ def build_options_schema(
         {
             vol.Optional(
                 CONF_HA_SENSOR_INDOOR_TEMPERATURE,
-                default=current_indoor_temp,
                 description={"suggested_value": current_indoor_temp},
             ): selector.EntitySelector(
                 selector.EntitySelectorConfig(
@@ -55,7 +54,6 @@ def build_options_schema(
             ),
             vol.Optional(
                 CONF_HA_SENSOR_CURRENT_POWER_CONSUMPTION,
-                default=current_power_consumption_sensor,
                 description={"suggested_value": current_power_consumption_sensor},
             ): selector.EntitySelector(
                 selector.EntitySelectorConfig(domain="sensor", device_class="power")
--- a/tests/test_schema_helper.py
+++ b/tests/test_schema_helper.py
@@ -12,7 +12,11 @@ import voluptuous as vol
 import voluptuous_serialize
 
 from custom_components.luxtronik2.const import (
+    CONF_HA_SENSOR_CURRENT_POWER_CONSUMPTION,
+    CONF_HA_SENSOR_INDOOR_TEMPERATURE,
+    CONF_UPDATE_INTERVAL,
     DEFAULT_PORT,
+    DEFAULT_UPDATE_INTERVAL_OPTION,
 )
 from custom_components.luxtronik2.schema_helper import (
     build_options_schema,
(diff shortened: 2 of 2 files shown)


rhammen:
Thank you for reporting this, and doubly so for coming back with a fix — the issue report alone would have been useful, but you traced it to the exact line and got the root cause right.

`vol.Optional(key, default=None)` becomes `default_factory=lambda: None`, so the key is forced into the validated output as `None` and `EntitySelector.__call__` rejects it via `cv.entity_id_or_uuid`. Dropping `default=` and leaning on `description={"suggested_value": ...}` for the UI pre-fill is exactly the right call — that's a well-targeted two-line change, and the write-up made it easy to verify. 🙏

I've pushed a regression test onto your branch so the behaviour stays locked in — hope that's alright, shout if you'd rather I had opened a separate PR.

### 🧪 What the test covers

Four cases in `TestBuildOptionsSchema`, all of which fail without your change and pass with it:

1. **Empty submission** — the bug you reported: the schema must validate when both sensor fields are left empty.
2. **Only one of the two sensors set** — pre-fix the error fired on *whichever* field was empty, so you couldn't set one sensor without also setting the other.
3. + 4. **Clearing a previously-set sensor** (parametrized over both fields) — see below.

That third case is why I wanted a test in place: while verifying your fix I found it also resolves a **second, quieter bug**. Clearing an already-configured sensor was silently impossible. The frontend omits emptied optional keys from the submission, and `default=current_value` then re-inserted the **old** value — so the field looked cleared in the UI but saved the previous sensor straight back.

Running the real `build_options_schema()` before vs after:

| scenario | before | after |
|---|---|---|
| both never set, left empty | ❌ error | `temp=None  power=None` |
| both set, **both** cleared | old values restored | `temp=None  power=None` |
| both set, only **power** cleared | old values restored | `temp=sensor.t  power=None` |
| both set, only **temp** cleared | old values restored | `temp=None  power=sensor.p` |
| power newly picked, temp empty | ❌ error | `power=sensor.p2` |

Without a test guarding it, re-adding `default=` later would quietly bring both bugs back, and the clearing one in particular is the kind that gets reported months later as "the setting won't stick".

### ✅ Downstream checks out

The explicit `None` that `config_flow.py` stores matters — both consumers read `entry.options.get(key, entry.data.get(key))`, so a present-but-`None` option correctly shadows a stale value in `entry.data`. `climate.py` guards with `is not None`, and `sensor.py`'s COP sensor falls back to the heat pump's internal `denominator_key`. Saving options triggers `async_reload`, so a cleared value takes effect immediately without a restart.

Full suite: 961 passed, coverage still 100% on `schema_helper.py` and `config_flow.py`. basedpyright 0 errors, ruff + codespell clean.

One residual edge, theoretical and not blocking: if the frontend ever sent the key with an explicit `None` instead of omitting it, validation would still fail the same way. Standard `ha-form` behavior is to delete empty optional keys, and your testing on HA 2026.8.1 confirms it in practice.

Thanks again — nice catch. 🎉

rhammen:
@mikeborgeld Merged it, thanks again for your first contribution!

mikeborgeld:
@rhammen Thank you for your detailed response and amazing effort. As this is my first ever GitHub contribution, it is a nice learning experience. I'll keep an eye out for another opportunity to contribute.

PR #774, opened 2026-08-27
fix(coordinator): bridge no_request dips while cooling with a bounded…

### LINKED ISSUE #773: When cooling is active status flips between cooling and no_request
Hi,

While cooling, `sensor.luxtronik_281211_0164_status` flaps between `cooling` and `no_request` on almost every 60s poll cycle, even though the heat pump is continuously cooling (temperatures and the flow-out target confirm this). Proposal: add an extra rescue path to the existing status derivation - detect an active cooling request via `ID_WEB_Sollwert_TRL_HZ` (C0012, `sensor.luxtronik_281211_0164_flow_out_temperature_target`), which stays pinned at a fixed value for the entire cooling cycle and moves back on the exact minute the cycle ends - with `ID_Einst_BA_Kuehl_akt` (P0108, `switch.luxtronik_281211_0164_cooling`) as a hard veto: when it is `off`, the status must never come out as `cooling`.

## Problem

During cooling, the status sensor drops to `no_request` for 1-3 minutes at a time, every few minutes, then returns to `cooling`. Example from my history (60s polling, every dip lasts exactly one or a few polls):

```
cooling     2026-08-26T15:53:08
no_request  2026-08-26T15:55:08
cooling     2026-08-26T15:56:08
no_request  2026-08-26T15:57:08
cooling     2026-08-26T15:58:08
...
```

I use the status to switch a utility-meter tariff and track consumption per operating mode (still more accurate for me than the energy registers added in a recent update), so every dip mis-attributes energy.

## What I found (some AI-assisted digging in `common.py` / `_derive_operation_mode`)

- The raw status word and the display-line registers are what flap; the workarounds in `_derive_operation_mode()` can only rescue a `no_request` poll when status line 3 shows `cooling` or the temperature heuristic from #404 passes, and during these dips neither holds.
- `binary_sensor..._approval_cooling` (`ID_WEB_FreigabKuehl`, C0146, binary_sensor.luxtronik_281211_0164_approval_cooling) is permanently `off` on my unit, even while it is demonstrably cooling — so it cannot serve as evidence here.
- `sensor..._flow_out_temperature_target` (`ID_WEB_Sollwert_TRL_HZ`, C0012, `sensor.luxtronik_281211_0164_flow_out_temperature_target`) turns out to be a rock-solid cooling indicator: it goes to a fixed 5.0 °C the minute a cooling cycle starts, stays there **continuously across every single dip**, and returns to its idle/heating value (15 °C at my place, dynamic in heating season) on the exact minute the cycle ends. Two full cycles in my history match the status boundaries to the minute (10:28–11:45 and 13:57–01:43).
- Note the 5.0 is not one of my configured cooling parameters (cooling target MK2 = 18 °C, cooling min flow-out = 17 °C, cooling threshold = 20 °C), so it looks like a fixed controller value rather than a setpoint mirror — comparing against the configured setpoints would not work, at least not on my unit.

## Proposal

Add an extra detection path to the existing workarounds in `_derive_operation_mode()` - not replacing anything, the current #118/#404 paths stay as they are:

1. New guard on all cooling outcomes: when the cooling mode switch (`ID_Einst_BA_Kuehl_akt` P0108, `switch.luxtronik_281211_0164_cooling`) is `off`, the heat pump is not allowed to cool at all, so the derived status should never come out as `cooling` - whatever the other registers show. This also hardens the existing #118/#404 paths against false positives.
2. New rescue path (alongside 'status line 3 = cooling' and the #404 temperature heuristic): when the raw status says `no_request` while that switch is `on` and `ID_WEB_Sollwert_TRL_HZ` (C0012, `sensor.luxtronik_281211_0164_flow_out_temperature_target`) sits at its cooling value - 5.0 °C on my unit - report `cooling`. In my history this register is pinned at 5.0 for the entire cooling cycle, straight through every `no_request` dip, and moves back to its idle/heating value on the exact minute the cycle ends. It tracks the real cooling request one-to-one, with no flapping and no delay in either direction.

One implementation note: the 5.0 does not match any of my configurable cooling parameters (cooling target MK2 = 18 °C, cooling min flow-out = 17 °C, cooling threshold = 20 °C), so it may be a fixed controller value - but I can't verify it is the same on every unit. If it can't be hardcoded, it could be latched from a poll that genuinely reports `cooling` and compared against on the polls where the status word dips to `no_request`.

Heat pump: Alpha Innotec SWCV92K3 FW: V3.92.1.

[history (5).csv](https://github.com/user-attachments/files/31509051/history.5.csv)
Image

onsmam:
… transition hold

Fixes #773

## 🔍 What this fixes

During (passive) cooling the status sensor flaps `cooling` ↔ `no_request` on nearly every 60s poll — see #773 for history exports and a diagnostics-confirmed live capture. Root cause: the raw status word (C0080) dips to `no request` while status line 3 shows `heating`, so only the #404 temperature heuristic can rescue the poll, and during passive cooling the flow/return delta it depends on hovers around zero whenever the compressor pauses. On the diagnosed unit the heat-source-in register (calc 204) additionally reads 0.0 (not connected), so the whole heuristic hangs on that one delta. Meanwhile `ID_WEB_Sollwert_TRL_HZ` (C0012) stays pinned at a fixed value (5.0 °C on this unit) for the entire cooling cycle — straight through every dip — and moves back on the exact minute the cycle ends.

## ✨ Changes

- `coordinator.py`: new `_update_cooling_transition_hold()` + `_cooling_evidence()`, mirroring the DHW transition hold from #728. Every poll that genuinely reports `cooling` latches the current C0012 value and arms a bounded hold window (5 min, re-armed each cooling poll). A `no_request` dip inside that window keeps reporting `cooling` while C0012 is still at the latched value (±0.1); the moment it moves, the hold releases immediately — no tail. Fallback evidence when C0012 is unavailable: cooling approval (C0146), heat-source pump (C0043) or heat-source flow rate (C0173). The time cap stays as safety net for units whose C0012 does not move between cooling and idle.
- `common.py`: the hold is honoured in `normalize_sensor_value()`, scoped to `no_request` so a genuine switch to heating/DHW/EVU is never masked (DHW hold takes precedence). New `_veto_cooling_when_switched_off()`: with the cooling mode switch (P0108) off the derived status can never come out as `cooling` at all — this also hardens the existing #118/#404 paths against false positives. A unit that does not expose P0108 (None) is not vetoed.
- `model.py`: new per-poll `cooling_transition_hold` flag on `LuxtronikCoordinatorData` (defaulted, so other construction sites are unaffected).
- `const.py`: new `COOLING_TRANSITION_HOLD` constant with rationale.

## 🧪 Tests

- Verified on real hardware (Alpha Innotec SWCV92k3, V3.92.1) with two HA instances reading the same heat pump: stock 2026.08.18 kept flapping, this branch reported one uninterrupted `cooling` block and returned to `no_request` within one poll of the cycle end.
- 18 new unit tests (coordinator hold lifecycle + status derivation incl. the P0108 veto); full suite passes (1188), ruff clean.

[history (6).csv](https://github.com/user-attachments/files/31514794/history.6.csv) --> Old version
[history (7).csv](https://github.com/user-attachments/files/31514787/history.7.csv) --> New version


rhammen:
Closing this one: After analysis this PR was superseded by #775, using a simpler approach, see discussion in #773

